PR #173, opened 2021-04-06
chore: Change log generation update

richard-bogdan-bitrise:
Change logs are now generated only for the modules that are defined in the input.

APM-2537


richard-bogdan-bitrise:
Good idea on the logging, added it!

richard-bogdan-bitrise:
Rebased.

PR #176, opened 2021-04-07
fix: remove key description from startup latency metric

daniellevass-bitrise:
Update the label_key description in the startup latency metric - the task said to make this an empty string, however none of the other metrics have a key description so i figured removing it entirely made sense :-)

I also had to update the metric_request.json test and the MetricConverter test.


richard-bogdan-bitrise (approved):
Excellent! 👍 

Just for to be more cautious, maybe you could ask for a thumbs up from the backend team? But this is just a recommendation, feel free to merge it.

daniellevass-bitrise:
> Excellent! 👍
> 
> Just for to be more cautious, maybe you could ask for a thumbs up from the backend team? But this is just a recommendation, feel free to merge it.

I was going to ping so they're aware the golden metrics file is changed :-)

PR #177, opened 2021-04-07
fix: Update metric endpoint

daniellevass-bitrise:
Updated the metric endpoint and updated the new headers that are required now for this one endpoint.

Tested by making a local trace-sdk and using the pineapple app. I received the following response:
Screenshot 2021-04-07 at 13 11 23

### DIFF
--- a/trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java
@@ -216,14 +216,12 @@ public final class NetworkClient {
         return chain -> {
             final Request request =
                     chain.request().newBuilder()
-                         .addHeader("Content-Type", "application/json")
-                         .addHeader("Accept-Language", getAcceptedLanguage())
-                         .addHeader("Connection", "keep-alive")
-                         .addHeader("Accept", "application/json")
-                         .addHeader("User-Agent", getUserAgent())
-                         .addHeader("Content-Length", getContentLength(chain))
                          .addHeader("Authorization", getAuthorizationBearer())
                          .addHeader("Accept-Encoding", "gzip;q=1.0, compress;q=0.5)")
+                         .addHeader("Accept-Language", getAcceptedLanguage())
+                         .addHeader("Connection", "keep-alive")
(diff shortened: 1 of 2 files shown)


richard-bogdan-bitrise:
Thanks for adding the response, can you please check the request too?

So we set the headers originally here:
[https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java#L218](https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java#L218)

and according to the javadoc of Headers annotation, headers with the same name are not overwritten, but a new one will be added:
[https://square.github.io/retrofit/2.x/retrofit/](https://square.github.io/retrofit/2.x/retrofit/)

To avoid confusions, I would remove these two header fields from NetworkClient, and manually add to sendTraces in NetworkCommunicator the Accept and Content type fields. So these two fields will be in the same class, both for metrics and traces. Might consider to add this information to the javadocs. Please let me know your thoughts about this.

richard-bogdan-bitrise (approved):

daniellevass-bitrise:
> Thanks for adding the response, can you please check the request too?
> 
> So we set the headers originally here:
> https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java#L218
> 
> and according to the javadoc of Headers annotation, headers with the same name are not overwritten, but a new one will be added:
> https://square.github.io/retrofit/2.x/retrofit/
> 
> To avoid confusions, I would remove these two header fields from NetworkClient, and manually add to sendTraces in NetworkCommunicator the Accept and Content type fields. So these two fields will be in the same class, both for metrics and traces. Might consider to add this information to the javadocs. Please let me know your thoughts about this.

Thanks for the feedback Richard!

I removed the headers from the NetworkClient (and alphabeticalised the list) and moved them into the NetworkConnector. I also added some javadoc to explain which was the updated endpoint, and that hopefully we can make the traces endpoint updated at some point. Does that look okay to you?

PR #178, opened 2021-04-07
chore: add codecov step to build trace-sdk

daniellevass-bitrise:
Updates the build-trace-sdk workflow to run unit tests with code coverage.
Updates the test_trace-android-sdk workflow to send the code coverage reports.

This means both unit tests, and integration tests reports are sent and combined by codecov. Currently, they are sent 4 times as we run the test_trace-android-sdk on 4 different sdk versions. However, this shouldn't really be an issue as codecov combines all the received reports. 

You can see an example report here: https://codecov.io/gh/bitrise-io/trace-android-sdk/commit/0d7fb2ca030a15559704671dc72a0cc309f3a669/

Once this has been merged. I would like to turn on the GitHub Webhook which means it will post into each open PR the codecov status (whether it's been increased or decreased).


trace-sdk/build.gradle
```diff
@@ -9,6 +10,7 @@ android {
         targetSdkVersion 30
         versionCode = "$project.versionCode" as int
         versionName "$project.version"
+        buildConfigField('String', 'VERSION_NAME', "\"${project.version}\"")
```
daniellevass-bitrise, line 13: this has to be added to get access to the VERSION_NAME in the NetworkClient - as per https://issuetracker.google.com/issues/158695880 :(

build.gradle
```diff
@@ -18,7 +18,8 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.2'
+        classpath 'com.android.tools.build:gradle:4.1.3'
+        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.16.0"
```
daniellevass-bitrise, line 22: I'm using [this library](https://github.com/vanniktech/gradle-android-junit-jacoco-plugin) instead of the official jacoco library because it currently has [this issue](https://github.com/gradle/gradle/issues/14132) :( There is an [alternative library](https://github.com/arturdm/jacoco-android-gradle-plugin/i) which codecov recommended however it has not been updated in a year and consequently also has[ compatability issues](https://github.com/arturdm/jacoco-android-gradle-plugin/issues/83)

build.gradle
```diff
@@ -18,7 +18,8 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.2'
+        classpath 'com.android.tools.build:gradle:4.1.3'
```
daniellevass-bitrise, line 21: There was a bug in the current android gradle build tools which meant that the jacoco reports never waited for the android connected tests to actually run, making them always fail the build. Updating fixed the issue, but it also meant [this was needed to update too](https://github.com/bitrise-io/trace-android-sdk/pull/178/files#r608816417)

build.gradle
richard-bogdan-bitrise: Thanks for the details, it was super useful. I checked the issue, and I was thinking it should not be the end of the world, because it is about the order of the task execution. A simple fix, with making the Jacoco task to depend on the connectedAndroidTask should fix it. Turned out, they made the connectedAndroidTask to depend on Jacoco task, so a single line of fix is not enough, as it would lead to circular task dependency (Jacoco task depends on connectedAndroidTest, connectedAndroidTest depends on Jacoco task).
So additionally, I had to update the task dependency of connectedAndroidTest, to not to depend on Jacoco task. Created an experimental branch for it, see the related [commit](https://github.com/bitrise-io/trace-android-sdk/commit/3b1926a547f637f655332678cfe476884e6dc1f1). The implementation of the official plugin is based on [this article](https://medium.com/wandera-engineering/android-kotlin-code-coverage-with-jacoco-sonar-and-gradle-plugin-6-x-3933ed503a6e), I just tailored it to our needs (removed some code, which are not relevant for us) and added the possible workaround/fix.

I do not want to rewrite your PR, but I was thinking we could stick with the official plugin, as I usually try to avoid 3rd party plugins. This is not a blocker, so happy to approve your PR as is, and move on, but left here for consideration (and review the experiment). 🙂

richard-bogdan-bitrise (approved):

PR #180, opened 2021-04-09
Experiment to resolve Gradle issue #14132 with code cov.

richard-bogdan-bitrise:


trace-sdk/build.gradle
```diff
@@ -32,6 +33,92 @@ android {
...
+        // Filter out Jacoco report tasks. this is an issue, connected tests should not depend on them
+        String providerName = ((TaskProvider) it).getName()
+        if (!reportTaskNames.grep{it == providerName}) {
+            deps.add(it)
+        }
+    }
+    // Replace the original dependsOn with the filtered
+    connectedAndroidTestTask.setDependsOn(deps)
+
+    // Finally make sure that each report task depends on the connected tests.
```
daniellevass-bitrise, line 112: this looks like it's failing the build as it's trying to run the connected tests as well i'm not sure this is accurate - we run the unit tests in the build-trace_sdk workflow and this now fails.

bitrise.yml
```diff
@@ -293,10 +293,6 @@ workflows:
             - module: trace-sdk
             - arguments: "--stacktrace"
           title: android_build
-      - gradle-runner@1:
```
daniellevass-bitrise, line 296: not sure removing the unit tests here is the best idea - possibly just revert the previous workflow step and run the unit tests without coverage? I presume the unit tests then get run later with the coverage at the same time the instrumentation tests run?

Badlazzor:
Closing as part of a stale-backlog cleanup pass — feel free to re-open if still relevant.

PR #182, opened 2021-04-13
feat: Support Android version 21

daniellevass-bitrise:
Updated the minsdk for the trace-sdk and trace-test-application.
Updated the test-sdk-22 workflow to run on 21 instead, as that's our new minimum.

There are some issues with the UI tests on sdk21 - specifically the uiautomator tests cannot "find" ui elements by resource id. ~~I updated the BySelector to match what's written in the documentation (changing it to find by PACKAGE_NAME - https://developer.android.com/training/testing/ui-testing/uiautomator-testing#verifying-results however this didn't fix the issue. I can revert this change if you'd prefer to keep it as it was.~~

I spent half a day trying to make the two tests work on sdk21 but made little progress, so have annotated those specific tests to not run on sdk21. Perhaps this could be a timeboxed subtask for APM-2110 (improving the test ui reliability)?

~~Updated the tests to work for sdk21 as there was an issue "finding" ui elements.~~

Note - the sdk21 tests are flaky as the emulator sometimes dies, however this was already a problem with test-sdk22 workflow as well.


richard-bogdan-bitrise:
> There are some issues with the UI tests on sdk21 - specifically the uiautomator tests cannot "find" ui elements by resource id. I updated the BySelector to match what's written in the documentation (changing it to find by PACKAGE_NAME) however this didn't fix the issue. I can revert this change if you'd prefer to keep it as it was.

If there is no fix or improvement with it, I think the best is to keep it as is, as that solution worked before.

richard-bogdan-bitrise:
> I spent half a day trying to make the two tests work on sdk21 but made little progress, so have annotated those specific tests to not run on sdk21. Perhaps this could be a timeboxed subtask for APM-2110 (improving the test ui reliability)?

What was the issue/error message? I would be more comfortable if it worked, or if we cannot find a solution, create some new test cases, that cover these, maybe with a different tool/framework, like Espresso. Without these tests we lose some life belt that would warn us about potential issues.

trace-test-application/src/androidTest/java/io/bitrise/trace/testapp/screen/IndexActivityScreen.java
```diff
@@ -34,7 +31,16 @@ public IndexActivityScreen(@NonNull final UiDevice uiDevice) {
      */
     @NonNull
     public MainActivityScreen launchUiTests() {
-        click(uiTestsButton);
+        UiObject button = uiDevice.findObject(new UiSelector()
```
daniellevass-bitrise, line 34: the change here is i followed the documentation to access ui components - https://developer.android.com/training/testing/ui-testing/uiautomator-testing#accessing-ui-components - i could refactor this out but it's only used in two places? What do you think?

richard-bogdan-bitrise: Something is not ok with it, as now the CI fails on all API levels. Could you please check it?

richard-bogdan-bitrise: Or, as Colin approved it, you can revert, suppress the running below API level 22, and if the CI passes we can merge it.

daniellevass-bitrise: I'm aware it failed - but it failed on the next part which demonstrated that getting element by text does work - so i was trying to update all the "click" to pass a UiSelector, not a BySelector. I expect it to fail a few more times until i catch all of them.

daniellevass-bitrise: So, I think i was going down the right track with getting elements by text, nor resource id - but it'd involve changing too much for this PR. I'm not wholly confident that this change actually worked either - the tests passed but the screenshots did not so i think it was a false positive (see the test-sdk21 workflow build) :(

I've created https://bitrise.atlassian.net/browse/APM-2795 to track the work needed to fix - we can discuss there the proposed way forward?

richard-bogdan-bitrise (approved):

trace-test-application/src/androidTest/java/io/bitrise/trace/testapp/screen/IndexActivityScreen.java
richard-bogdan-bitrise: Sure, let's discuss it in the mentioned ticket.

PR #185, opened 2021-04-19
chore: add flag to TraceSdk to increase debug logging

daniellevass-bitrise:
This PR adds a flag `DEBUG_ENABLED` to allow customers to use TraceSdk in debug mode. I've also turned it on in the trace-test-application.

I've added some logging to tell people what state debug mode is at the start, and when it gets changed. That might look confusing to people who override it in their first activity - What do you think?

Screenshot 2021-04-19 at 14 30 34
> example of the logs when you start the SDK

### DIFF
--- a/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
@@ -41,6 +41,11 @@ public class TraceSdk {
      */
     public static final String NAME = "Trace Android";
 
+    /**
+     * Flag to configure the sdk into debug mode, by default this will be false.
+     */
+    public static boolean DEBUG_ENABLED = false;
+
     private TraceSdk() {
         // nop
     }
@@ -69,9 +74,19 @@ public class TraceSdk {
             initDataCollection(context);
             initLifeCycleListener(context);
             initNetworkTracing();
+            TraceLog.i(String.format(LogMessageConstants.TRACE_DEBUG_FLAG_STATUS, DEBUG_ENABLED));
         } else {
             TraceLog.e(new TraceException.TraceConfigNotInitialisedException());
         }
+    }
+
+    /**
+     * Flag to enable TraceSdk into debug mode.
+     * @param debugEnabled boolean value to enable or disable debug mode in the TraceSdk.
+     */
+    public synchronized static void setDebugEnabled(boolean debugEnabled) {
+        DEBUG_ENABLED = debugEnabled;
+        TraceLog.i(String.format(LogMessageConstants.TRACE_DEBUG_FLAG_STATUS, DEBUG_ENABLED));
     }
 
     /**
--- a/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
@@ -43,7 +43,8 @@ public class TraceSdk {
 
     /**
      * Flag to configure the sdk into debug mode, by default this will be false.
-     * Currently, debug mode will add more debug log messages.
+     * Currently, debug mode will add more debug log messages. Note: we do not log anything if your
+     * app is in release mode. To see these log messages your app should also be in debug mode.
      */
     private static boolean DEBUG_ENABLED = false;
 
@@ -83,6 +84,8 @@ public class TraceSdk {
 
     /**
      * Flag to enable TraceSdk into debug mode. Currently, this will add more debug log messages.
+     * Note: we do not log anything if your app is in release mode. To see these log messages your
+     * app should also be in debug mode.
      *
      * @param debugEnabled boolean value to enable or disable debug mode in the TraceSdk.
      */
--- a/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
@@ -42,9 +42,12 @@ public class TraceSdk {
     public static final String NAME = "Trace Android";
 
     /**
-     * Flag to configure the sdk into debug mode, by default this will be false.
-     * Currently, debug mode will add more debug log messages. Note: we do not log anything if your
-     * app is in release mode. To see these log messages your app should also be in debug mode.
+     * The TraceSdk has a debug mode - currently this will mean more debug level log messages.
+     *
+     * Please note if you are not using a debug build, and or minify is enabled it can affect these
+     * logs, and they can be stripped out depending on your configuration.
+     * You also need to ensure that the TraceSdk has been initialised before setting the debug
+     * enabled mode.
      */
     private static boolean DEBUG_ENABLED = false;
 
@@ -83,9 +86,12 @@ public class TraceSdk {
     }
 
     /**
-     * Flag to enable TraceSdk into debug mode. Currently, this will add more debug log messages.
-     * Note: we do not log anything if your app is in release mode. To see these log messages your
-     * app should also be in debug mode.
+     * Flag to enable debug mode - currently this will mean more debug level log messages.
+     *
+     * Please note if you are not using a debug build, and or minify is enabled it can affect these
+     * logs, and they can be stripped out depending on your configuration.
+     * You also need to ensure that the TraceSdk has been initialised before setting the debug
+     * enabled mode.
      *
      * @param debugEnabled boolean value to enable or disable debug mode in the TraceSdk.
      */
--- a/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
@@ -43,8 +43,9 @@ public class TraceSdk {
 
     /**
      * Flag to configure the sdk into debug mode, by default this will be false.
+     * Currently, debug mode will add more debug log messages.
      */
-    public static boolean DEBUG_ENABLED = false;
+    private static boolean DEBUG_ENABLED = false;
 
     private TraceSdk() {
         // nop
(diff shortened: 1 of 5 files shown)


trace-test-application/build.gradle
```diff
@@ -36,7 +36,7 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation "io.bitrise.trace:trace-sdk:${traceVersions.traceSdk}"
+    implementation project(path: ':trace-sdk')
```
daniellevass-bitrise, line 39: this was an Android Studio suggestion because it couldn't pick up the TraceSdk change without publishing to mavenLocal. I don't think it should cause any issues - and it should make things faster to test locally :-)

richard-bogdan-bitrise:
Have you considered using the debug build type for this purpose, or creating a new variant? That way we could 
- easily put the required code to a different sourceset
- making it easier to understand and more readable
- developers could easily configure this behaviour in their build.gradle by adding the proper variant dependency of the sdk

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
```diff
@@ -69,11 +74,21 @@ public synchronized static void init(@NonNull final Context context) {
...
             initLifeCycleListener(context);
             initNetworkTracing();
+            TraceLog.i(String.format(LogMessageConstants.TRACE_DEBUG_FLAG_STATUS, DEBUG_ENABLED));
         } else {
             TraceLog.e(new TraceException.TraceConfigNotInitialisedException());
         }
     }
 
+    /**
+     * Flag to enable TraceSdk into debug mode.
```
richard-bogdan-bitrise, line 84: Could you please add more information on what does this debug mode mean? Developers might have questions what does this explicitly do.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
```diff
@@ -41,6 +41,11 @@
      */
     public static final String NAME = "Trace Android";
 
+    /**
+     * Flag to configure the sdk into debug mode, by default this will be false.
+     */
+    public static boolean DEBUG_ENABLED = false;
```
richard-bogdan-bitrise, line 47: I would recommend making this private, and access/mutate only with the getter/setter.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
```diff
@@ -69,11 +74,21 @@ public synchronized static void init(@NonNull final Context context) {
...
         } else {
             TraceLog.e(new TraceException.TraceConfigNotInitialisedException());
         }
     }
 
+    /**
+     * Flag to enable TraceSdk into debug mode.
+     * @param debugEnabled boolean value to enable or disable debug mode in the TraceSdk.
+     */
+    public synchronized static void setDebugEnabled(boolean debugEnabled) {
```
richard-bogdan-bitrise, line 87: nits: 
- please use the code formatter to add the missing extra line before param
- argument `debugEnabled` can be final

trace-sdk/src/main/java/io/bitrise/trace/utils/log/LogMessageConstants.java
```diff
@@ -50,6 +50,7 @@
     public static final String SET_URL_STREAM_HANDLER_FACTORY_FAILED = "setURLStreamHandlerFactory failed. This means" +
             " the Trace SDK will not be able to capture network metrics that use URLConnection. Please refer to the " +
             "documentation for further guidance.";
+    public static final String TRACE_DEBUG_FLAG_STATUS = "TraceSdk debug = %b";
```
richard-bogdan-bitrise, line 53: nit: I would make this message more verbose:  if I only see this in the logs I may not now what would it mean.

richard-bogdan-bitrise:
Nothing critical, some small fixes, and I would argue to work with different build type/variant instead of adding a flag.

trace-sdk/src/main/java/io/bitrise/trace/utils/log/LogMessageConstants.java
daniellevass-bitrise: Yeah I agree - but its really hard 😅 

Have updated it to read `TraceSdk debug mode enabled = %b` - does that sound any better? 

Would it be better if it said it was in production/debug mode ? 

I think we should also be updating the readme here to include what it means - i'll add that into this PR.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
daniellevass-bitrise: Have added a sentence on what it adds currently - ` Currently, this will add more debug log messages.` - i've deliberately left it light on exactly what kind of debug messages because i can see it being useful to add more in the future. Does that make sense?

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
daniellevass-bitrise: Have updated - thank you.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
daniellevass-bitrise: Have updated - thank you.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
```diff
@@ -69,11 +75,29 @@ public synchronized static void init(@NonNull final Context context) {
...
     }
 
+    /**
+     * Flag to enable TraceSdk into debug mode. Currently, this will add more debug log messages.
+     *
+     * @param debugEnabled boolean value to enable or disable debug mode in the TraceSdk.
+     */
+    public synchronized static void setDebugEnabled(final boolean debugEnabled) {
+        DEBUG_ENABLED = debugEnabled;
+        TraceLog.i(String.format(LogMessageConstants.TRACE_DEBUG_FLAG_STATUS, DEBUG_ENABLED));
```
richard-bogdan-bitrise, line 91: This will make logging silent implicitly, if the logger is not initialised yet. See [here](https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/main/java/io/bitrise/trace/utils/log/TraceLog.java#L128). I think we have to mention this in the readme, to set this enabled after the SDK is initialised, otherwise log messages will be silent.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
richard-bogdan-bitrise: Sure, mentioning it adds logging is enough, did not mean to add a list of messages, just what differences it has with the release mode.

trace-sdk/src/main/java/io/bitrise/trace/utils/log/LogMessageConstants.java
richard-bogdan-bitrise: I think this simple modification in the wording helps a lot in the understanding, thanks!

richard-bogdan-bitrise:
One note added, I would at least notify users about this behaviour in the README, otherwise the rest is ok.

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
daniellevass-bitrise: I added the following to the three places people will see info about it (the flag variable, the setter, and the readme):
> Note: we do not log anything if your app is in release mode. To see these log messages your app should also be in debug mode.

Does that look okay?

richard-bogdan-bitrise: I am not sure if that is correct, because developers can have such configuration, that their release version of the app does not strip out debug level logs with proguard.

But what I mentioned above in the comment is a different thing. If we call setDebugEnabled() method before the TraceSDK.init() is called, TraceLog will create an instance of SilentLogger, so no logs will be seen on Logcat.

daniellevass-bitrise: Hmm. Perhaps both are valid to mention?

How does this sound (figured i'd get your feedback before pushing the code):
> The TraceSdk has a debug mode - currently this will mean more debug level log messages. 
We do not log anything if your app is not in debug mode unless you have customised this with something like proguard. 
You also need to ensure that the TraceSdk has been initialised before setting the debug enabled mode.

richard-bogdan-bitrise: > We do not log anything if your app is not in debug mode unless you have customised this with something like proguard.

This depends on the state of minify, and of course the configuration of the rules. I believe the simplest solution is always the best: Please note if minify is enabled it can affect these logs, and they can be stripped out depending on your configuration.

daniellevass-bitrise: What do you mean by the simplest solution is always best ?

> The TraceSdk has a debug mode - currently this will mean more debug level log messages.
Please note if you are not using a debug build, and or minify is enabled it can affect these logs, and they can be stripped out depending on your configuration
You also need to ensure that the TraceSdk has been initialised before setting the debug enabled mode.

Does this work?

richard-bogdan-bitrise: Exactly, thanks!

richard-bogdan-bitrise (approved):
Approved!

trace-sdk/src/main/java/io/bitrise/trace/TraceSdk.java
daniellevass-bitrise: Perfect - will update the code!

daniellevass-bitrise: Thank you for your help on this btw - sometimes it takes a few messages to get the wording right - these things are really important :)

PR #186, opened 2021-04-20
chore: Track and Log metric counts from header validation

daniellevass-bitrise:
This PR adds a header validation method into the MetricSender. When it a metric request is successfully completed, it will compare what the request metrics were, and what the response headers were. It will log the following messages:
1. if the header is not available - a comparison could not be completed.
2. headers match - sent x, received y.
3. headers do not match sent x, received y, sent keys ..., received keys ...

I also added a new method to the the TraceLog to use the DEBUG_ENABLED flag to log only if that is set. 

Tested this with the pineapple app:

with the debug flag enabled:
Screenshot 2021-04-20 at 15 58 53

with the debug flag not enabled:
Screenshot 2021-04-20 at 16 07 34


trace-sdk/src/main/java/io/bitrise/trace/network/MetricSender.java
```diff
@@ -92,6 +96,39 @@ public void onSuccess() {
         });
     }
 
+    @VisibleForTesting
+    protected static int headerComparison(@NonNull final MetricRequest request,
```
richard-bogdan-bitrise, line 100: Please consider splitting this method to smaller parts, to make it more readable and testable.

Is it required to make it protected? If not, I would recommend to make it more restrictive and use the package private visibility.

Javadoc is missing, but it would help developers to understand what it does without checking the content.

nit: when I read the name, it confuses a bit me, because I would expect what it does (verb instead of a noun). Could you please make this more descriptive, something like `countNonAcceptedMetrics` or something like that?

trace-sdk/src/main/java/io/bitrise/trace/utils/log/LogMessageConstants.java
```diff
@@ -38,6 +38,9 @@
     public static final String FAILED_TO_READ_SYSTEM_CPU_STATS = "Failed to read System CPU stats, Trace may not " +
             "report any CPU stats for this session.";
     public static final String INITIALISING_SDK = "Initialising the Trace SDK";
+    public static final String METRIC_HEADERS_MISSING = "Metric response headers missing, failed to complete header validation.";
+    public static final String METRIC_HEADERS_MATCH = "Metric response headers match: %d sent and %d received.";
```
richard-bogdan-bitrise, line 42: nit: I think the number will be always the same in this case, so it will be redundant.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderTest.java
```diff
@@ -87,4 +94,60 @@ public void needReschedule_shouldBeTrue_hasAll() {
 
         assertTrue(metricSender.isRescheduleNeeded());
     }
+
+    private final String headerAcceptedMetricsCount = "accepted-metrics-count";
```
richard-bogdan-bitrise, line 98: Maybe we could put these two variables in a place where it is accessible for this class, so we do not have to type them twice, and modify them twice if it changes.

richard-bogdan-bitrise:
Generally ok, some small updates are recommended, also added some nits.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderTest.java
daniellevass-bitrise: Have added the constants to the NetworkCommunicator because that's where the network calls are declared, so it made sense to put the header constants near there. Does that sound okay?

trace-sdk/src/main/java/io/bitrise/trace/network/MetricSender.java
daniellevass-bitrise: Thanks for the feedback!

I moved the logic to generate the list of keys to the MetricUtils (and wrote tests) - hopefully that makes it clearer to see what's happening now?

Updated to make it package private, added javadoc, and renamed to `countHeaderComparisonDifference` - i think the difference is better than "non accepted" because there is the scenario where the backend accepts more than we think we sent. I added a test for this case too. Does that make sense?

trace-sdk/src/main/java/io/bitrise/trace/utils/log/LogMessageConstants.java
daniellevass-bitrise: Updated to remove the second number - now it just contains how many were sent.

trace-sdk/src/main/java/io/bitrise/trace/network/MetricSender.java
richard-bogdan-bitrise: Thanks for the updates.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderTest.java
richard-bogdan-bitrise: I would argue, that the purpose of the network client is the handling of the communication, and the given Sender classes should contain these constants as well (as the validation happens there). But if you would keep it here, I am ok with it, so this is not a blocker.

richard-bogdan-bitrise (approved):
Approved! 🚀

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderTest.java
daniellevass-bitrise: I'll make this change in #187 so that i can update all usages in one place.

PR #187, opened 2021-04-21
tests: Add integration tests for metrics

daniellevass-bitrise:
Adds an integration test for each metric we send and confirms the metric count and metric labels match what we expect to receive from the backend. 

Have also added a debug configurationManager initialisation so i could create an instance without having the bitrise-addons-configuration.json - not sure how you feel about this? Do you think thee is a better / different way we could have done this? Not precious about the implementation - just wanted to be able to send requests that had a valid auth header 😄


trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+import retrofit2.Response;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Integration tests that send a metric to the backend and expect confirmation by headers.
+ */
+public class MetricSenderIntegrationTest {
+
+    private static final String TRACE_TOKEN = "70a14e32-97b1-4b6d-b4bc-ff592c780325";
```
richard-bogdan-bitrise, line 26: I think we should not share any token here, they should be private.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+import io.bitrise.trace.test.MetricTestProvider;
+import io.opencensus.proto.metrics.v1.Metric;
+import retrofit2.Response;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Integration tests that send a metric to the backend and expect confirmation by headers.
+ */
+public class MetricSenderIntegrationTest {
```
richard-bogdan-bitrise, line 24: These tests are validating the same dummy values against the backend, but I think the ticket was to collect real data. Please update them accordingly.

Note: due to their nature (firing up real network calls), we should annotate them with LargeTest annotation to indicate it takes more time to complete.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+     * label were accepted.
+     * @param metric the {@link Metric} to send.
+     * @throws IOException if there are any issues executing the network request.
+     */
+    private void sendMetricList(@NonNull final Metric metric) throws IOException {
+        final List<Metric> metricList = new ArrayList<>();
+        metricList.add(metric);
+        final MetricRequest request = new MetricRequest(DataTestUtils.getSampleResource("sessionId"), metricList);
+        final Response<Void> response = NetworkClient.getCommunicator().sendMetrics(request).execute();
+        assertEquals("1", response.headers().get("accepted-metrics-count"));
```
richard-bogdan-bitrise, line 49: Please see [this](https://github.com/bitrise-io/trace-android-sdk/pull/186/files#r620065516) comment.

richard-bogdan-bitrise (changes requested):
Please remove sensitive content and if I am not mistaken, we should update the tests.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
daniellevass-bitrise: Updated to use the NetworkCommunicator constants I created in #186

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
daniellevass-bitrise: Good point. I had to move these tests then to `androidTest` not `test`.

I have updated all the tests except the ApplicationStartUp to call the collector, then the formatter and send the formatted data across. 

There was an issue with the ApplicationStartUp in that, the sdk doesn't actually have any screens - I could have mocked all the calls but I think that might be over complicating things. Instead, I checked what data got created when it called `onDataCollected` - this then still gets passed through the formatter. Does that sound okay?

trace-sdk/build.gradle
```diff
@@ -81,4 +82,16 @@ junitJacoco {
...
+def getTraceToken() {
+    // try to get the environment variable in CI
+    def traceToken = findProperty("traceToken")
+    if (traceToken != null) {
+        return traceToken
+    }
+    // if we don't have a env var - get it from the local.properties
+    Properties properties = new Properties()
+    properties.load(project.rootProject.file('local.properties').newDataInputStream())
+    return properties.getProperty("traceToken");
```
richard-bogdan-bitrise, line 94: Shouldn't we throw some error/exception, if this returns null?

trace-sdk/src/androidTest/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Integration tests that send a metric to the backend and expect confirmation by headers.
+ */
+@LargeTest
+public class MetricSenderIntegrationTest {
+
+    @Before
+    public void setUp() {
+        ConfigurationManager.getDebugInstance(BuildConfig.TRACE_TOKEN);
```
richard-bogdan-bitrise, line 49: nit: as it is initialised always with the same key, it would be enough to init/reset once per test class, no need to do it before each test case.

trace-sdk/src/test/java/io/bitrise/trace/network/MetricSenderIntegrationTest.java
richard-bogdan-bitrise: Sure, thanks for the updates!

richard-bogdan-bitrise (approved):
Thanks for the updates. Added some more thoughts, but nothing blocker! 🚀

trace-sdk/build.gradle
daniellevass-bitrise: that's a good question - i figured it was okay to silently fail as it's only used in one place and it's likely most people aren't going to use it. If they run the test that needs it - the tests would fail. Perhaps it makes sense to add a note to those tests to make sure people have it set?

PR #189, opened 2021-04-28
tests: data formatter tests

daniellevass-bitrise:
Have added unit tests to cover all formatters.

Before:
Screenshot 2021-05-05 at 13 35 06

After:
Screenshot 2021-05-05 at 13 32 56

Not sure why codecov isn't including the unit test coverage :(

### DIFF
--- a/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
@@ -69,7 +69,7 @@ public class ApplicationCpuDataFormatter extends DataFormatter {
      * @return the created TimeSeries.
      */
     @NonNull
-    private static TimeSeries createCpuTimeSeriesEntry(@NonNull final Timestamp timestamp, final Double value) {
+    private static TimeSeries createCpuTimeSeriesEntry(@NonNull final Timestamp timestamp, final double value) {
         return TimeSeries.newBuilder()
                          .addPoints(
                                  Point.newBuilder()
--- a/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
@@ -32,12 +32,8 @@ public class ApplicationCpuDataFormatter extends DataFormatter {
             return new FormattedData[]{};
         }
         final Double appCpuStat = (Double) data.getContent();
-        if (appCpuStat == null) {
-            return new FormattedData[]{};
-        }
         final Timestamp timestamp = getTimestamp();
         final Metric appCpuMetric = createAppCpuMetric(appCpuStat, timestamp);
-
 
         return new FormattedData[]{new FormattedData(appCpuMetric)};
     }
@@ -59,7 +55,7 @@ public class ApplicationCpuDataFormatter extends DataFormatter {
                                 .setUnit(percent)
                                 .setType(MetricDescriptor.Type.GAUGE_DOUBLE);
 
-        final TimeSeries appTimeSeries = createCpuTimeSeriesEntry(timestamp, (float) cpuUsagePercent);
+        final TimeSeries appTimeSeries = createCpuTimeSeriesEntry(timestamp, cpuUsagePercent);
         builder.setMetricDescriptor(cpuDescriptorBuilder.build());
         builder.addTimeseries(appTimeSeries);
         return builder.build();
@@ -73,7 +69,7 @@ public class ApplicationCpuDataFormatter extends DataFormatter {
      * @return the created TimeSeries.
      */
     @NonNull
-    private static TimeSeries createCpuTimeSeriesEntry(@NonNull final Timestamp timestamp, final float value) {
+    private static TimeSeries createCpuTimeSeriesEntry(@NonNull final Timestamp timestamp, final Double value) {
         return TimeSeries.newBuilder()
                          .addPoints(
                                  Point.newBuilder()
--- a/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatter.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatter.java
@@ -47,9 +47,6 @@ public class SystemCpuDataFormatter extends DataFormatter {
         }
 
         final CpuStat systemCpuStat = (CpuStat) data.getContent();
-        if (systemCpuStat == null) {
-            return new FormattedData[]{};
-        }
         final Timestamp timestamp = getTimestamp();
         final Metric systemCpuMetric = createSystemCpuMetric(systemCpuStat, timestamp);
 
--- a/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
+++ b/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
@@ -15,18 +15,17 @@ import static org.junit.Assert.assertNotNull;
  */
 public class ApplicationVersionCodeDataFormatterTest extends BaseDataFormatterTest {
 
-    final String resourceLabel = "app.build";
-    final String versionCode = "123";
-
     @Test
     public void formatData() {
+        final String versionCode = "123";
+
         final Data data = new Data(ApplicationVersionCodeDataCollector.class);
         data.setContent(versionCode);
         final ApplicationVersionCodeDataFormatter formatter = new ApplicationVersionCodeDataFormatter();
         final FormattedData[] formattedData = formatter.formatData(data);
         assertEquals(1, formattedData.length);
         assertNotNull(formattedData[0].getResourceEntity());
-        assertEquals(resourceLabel, formattedData[0].getResourceEntity().getLabel());
+        assertEquals("app.build", formattedData[0].getResourceEntity().getLabel());
         assertEquals(versionCode, formattedData[0].getResourceEntity().getValue());
     }
 }
--- a/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
+++ b/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
@@ -1,0 +1,32 @@
+package io.bitrise.trace.data.management.formatter.application;
+
+import org.junit.Test;
+
+import io.bitrise.trace.data.collector.application.ApplicationVersionCodeDataCollector;
+import io.bitrise.trace.data.dto.Data;
+import io.bitrise.trace.data.dto.FormattedData;
+import io.bitrise.trace.data.management.formatter.BaseDataFormatterTest;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Unit tests for {@link ApplicationVersionCodeDataFormatter}.
+ */
+public class ApplicationVersionCodeDataFormatterTest extends BaseDataFormatterTest {
+
+    final String resourceLabel = "app.build";
+    final String versionCode = "1.2.3";
+
+    @Test
+    public void formatData() {
+        final Data data = new Data(ApplicationVersionCodeDataCollector.class);
+        data.setContent(versionCode);
+        final ApplicationVersionCodeDataFormatter formatter = new ApplicationVersionCodeDataFormatter();
+        final FormattedData[] formattedData = formatter.formatData(data);
+        assertEquals(1, formattedData.length);
+        assertNotNull(formattedData[0].getResourceEntity());
+        assertEquals(resourceLabel, formattedData[0].getResourceEntity().getLabel());
+        assertEquals(versionCode, formattedData[0].getResourceEntity().getValue());
+    }
+}
--- a/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionNameDataFormatterTest.java
+++ b/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionNameDataFormatterTest.java
@@ -2,7 +2,7 @@ package io.bitrise.trace.data.management.formatter.application;
 
 import org.junit.Test;
 
-import io.bitrise.trace.data.collector.application.ApplicationVersionCodeDataCollector;
+import io.bitrise.trace.data.collector.application.ApplicationVersionNameDataCollector;
 import io.bitrise.trace.data.dto.Data;
 import io.bitrise.trace.data.dto.FormattedData;
 import io.bitrise.trace.data.management.formatter.BaseDataFormatterTest;
@@ -11,22 +11,22 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 /**
- * Unit tests for {@link ApplicationVersionCodeDataFormatter}.
+ * Unit tests for {@link ApplicationVersionNameDataFormatter}.
  */
-public class ApplicationVersionCodeDataFormatterTest extends BaseDataFormatterTest {
+public class ApplicationVersionNameDataFormatterTest extends BaseDataFormatterTest {
 
-    final String resourceLabel = "app.build";
-    final String versionCode = "123";
+    final String resourceLabel = "app.version";
+    final String versionName = "1.2.3";
 
     @Test
     public void formatData() {
-        final Data data = new Data(ApplicationVersionCodeDataCollector.class);
-        data.setContent(versionCode);
-        final ApplicationVersionCodeDataFormatter formatter = new ApplicationVersionCodeDataFormatter();
+        final Data data = new Data(ApplicationVersionNameDataCollector.class);
+        data.setContent(versionName);
+        final ApplicationVersionNameDataFormatter formatter = new ApplicationVersionNameDataFormatter();
         final FormattedData[] formattedData = formatter.formatData(data);
         assertEquals(1, formattedData.length);
         assertNotNull(formattedData[0].getResourceEntity());
         assertEquals(resourceLabel, formattedData[0].getResourceEntity().getLabel());
-        assertEquals(versionCode, formattedData[0].getResourceEntity().getValue());
+        assertEquals(versionName, formattedData[0].getResourceEntity().getValue());
     }
 }
--- a/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionNameDataFormatterTest.java
+++ b/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionNameDataFormatterTest.java
@@ -15,18 +15,17 @@ import static org.junit.Assert.assertNotNull;
  */
 public class ApplicationVersionNameDataFormatterTest extends BaseDataFormatterTest {
 
-    final String resourceLabel = "app.version";
-    final String versionName = "1.2.3";
-
     @Test
     public void formatData() {
+        final String versionName = "1.2.3";
+
         final Data data = new Data(ApplicationVersionNameDataCollector.class);
         data.setContent(versionName);
         final ApplicationVersionNameDataFormatter formatter = new ApplicationVersionNameDataFormatter();
         final FormattedData[] formattedData = formatter.formatData(data);
         assertEquals(1, formattedData.length);
         assertNotNull(formattedData[0].getResourceEntity());
-        assertEquals(resourceLabel, formattedData[0].getResourceEntity().getLabel());
+        assertEquals("app.version", formattedData[0].getResourceEntity().getLabel());
         assertEquals(versionName, formattedData[0].getResourceEntity().getValue());
     }
 }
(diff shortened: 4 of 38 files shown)


trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
```diff
@@ -32,13 +32,9 @@
             return new FormattedData[]{};
         }
         final Double appCpuStat = (Double) data.getContent();
-        if (appCpuStat == null) {
```
daniellevass-bitrise, line 35: Pretty sure this isn't possible - and gets caught by L31.

trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
```diff
@@ -59,7 +55,7 @@ public static Metric createAppCpuMetric(final double cpuUsagePercent, @NonNull f
                                 .setUnit(percent)
                                 .setType(MetricDescriptor.Type.GAUGE_DOUBLE);
 
-        final TimeSeries appTimeSeries = createCpuTimeSeriesEntry(timestamp, (float) cpuUsagePercent);
```
daniellevass-bitrise, line 62: this caused an interesting side effect in the tests
Screenshot 2021-04-28 at 12 24 14

You can see i'm setting the value to be 12.3 - but because it's cast to a float it's then equal to 12.300000190734863. Was there a reason for this cast? L81 we're calling `setDoubleValue` so i figured it was safe to just remove.

trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatter.java
```diff
@@ -47,9 +47,6 @@
         }
 
         final CpuStat systemCpuStat = (CpuStat) data.getContent();
-        if (systemCpuStat == null) {
```
daniellevass-bitrise, line 50: likewise, L45 covers this scenario - if it's a null value, it's not a CpuStat and so it gets caught there.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/memory/ApplicationUsedMemoryDataFormatterTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void formatData_contentNull() {
+        final Data data = new Data(ApplicationUsedMemoryDataCollector.class);
+        data.setContent(null);
+        assertArrayEquals(new FormattedData[]{}, formatter.formatData(data));
+    }
+
+    @Test
+    public void handleMemoryFormatting() {
```
daniellevass-bitrise, line 31: for this test ApplicationUsedMemory and SystemMemory - we already have a MemoryFormatterTest that ensures the Metric created is correct - so these just needed to make sure they set the correct metric descriptor. Does that make sense?

trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
richard-bogdan-bitrise: I do not see a reason for the cast, maybe the signature was different and the cast was needed, but after a change it is not required.
On the other hand I believe in L72 it should be a primitive double instead of Double, as null values are not acceptable in this case.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
```diff
@@ -0,0 +1,32 @@
...
+import io.bitrise.trace.data.dto.FormattedData;
+import io.bitrise.trace.data.management.formatter.BaseDataFormatterTest;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Unit tests for {@link ApplicationVersionCodeDataFormatter}.
+ */
+public class ApplicationVersionCodeDataFormatterTest extends BaseDataFormatterTest {
```
richard-bogdan-bitrise, line 16: Should we add a case for null content?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionNameDataFormatterTest.java
```diff
@@ -0,0 +1,32 @@
...
+import io.bitrise.trace.data.dto.FormattedData;
+import io.bitrise.trace.data.management.formatter.BaseDataFormatterTest;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Unit tests for {@link ApplicationVersionNameDataFormatter}.
+ */
+public class ApplicationVersionNameDataFormatterTest extends BaseDataFormatterTest {
```
richard-bogdan-bitrise, line 16: Same comment as [above](https://github.com/bitrise-io/trace-android-sdk/pull/189/files#r627270636)

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatterTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unit tests for {@link ApplicationCpuDataFormatter}.
+ */
+public class ApplicationCpuDataFormatterTest extends BaseDataFormatterTest {
+
+    final ApplicationCpuDataFormatter formatter = new ApplicationCpuDataFormatter();
+    final Double cpuValue = 12.3;
+    final String metricDescriptorDescription = "Application CPU Usage";
+
+    @Test
+    public void formatData_contentNotDouble() {
```
richard-bogdan-bitrise, line 32: Same comment as [above](https://github.com/bitrise-io/trace-android-sdk/pull/189/files#r627270636)

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatterTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final Data data = new Data(ApplicationCpuUsageDataCollector.class);
+        data.setContent("cpu usage");
+        assertArrayEquals(new FormattedData[]{}, formatter.formatData(data));
+    }
+
+    @Test
+    public void formatData() {
```
richard-bogdan-bitrise, line 39: nit: I think it would make it more readable if we could split this up to smaller parts. I was thinking a new method something like getExpectedMetric would help here.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatterTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        metricBuilder.setMetricDescriptor(cpuDescriptorBuilder.build());
+        metricBuilder.addAllTimeseries(expectedTimeSeries);
+        final Metric expectedMetric = metricBuilder.build();
+
+        assertEquals(expectedMetric, formattedData.getMetricEntity().getMetric() );
+    }
+
+    @NonNull
+    private static TimeSeries createCpuTimeSeriesEntry(@NonNull final Timestamp timestamp,
```
richard-bogdan-bitrise, line 96: We do not have to copy this here, we just have to update the original method in SystemCpuDataFormatter to package default visibility (and add VisibleForTesting annotation). This way we could avoid code duplication.

Note: related to this [comment](https://github.com/bitrise-io/trace-android-sdk/pull/189/files#r622139103), I think the original method should be updated to use double as a param.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceCarrierDataFormatterTest.java
```diff
@@ -0,0 +1,29 @@
...
+import io.bitrise.trace.data.resource.ResourceEntity;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link DeviceCarrierDataFormatter}.
+ */
+public class DeviceCarrierDataFormatterTest extends BaseDataFormatterTest {
+
+    final String carrier = "vodafone";
```
richard-bogdan-bitrise, line 18: nit: private visibility is enough for this variable.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceIdDataFormatterTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.bitrise.trace.data.resource.ResourceEntity;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link DeviceIdDataFormatter}.
+ */
+public class DeviceIdDataFormatterTest extends BaseDataFormatterTest {
+
+    final String devideId = UUID.randomUUID().toString();
```
richard-bogdan-bitrise, line 20: nit: private visibility is enough for this variable.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceLocaleDataFormatterTest.java
```diff
@@ -15,6 +16,7 @@
  */
 public class DeviceLocaleDataFormatterTest extends BaseDataFormatterTest {
 
+    final String deviceLocale = "en_US";
```
richard-bogdan-bitrise, line 19: nit: private visibility is enough for this variable.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceModelDataFormatterTest.java
```diff
@@ -0,0 +1,29 @@
...
+import io.bitrise.trace.data.resource.ResourceEntity;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link DeviceModelDataFormatter}.
+ */
+public class DeviceModelDataFormatterTest extends BaseDataFormatterTest {
+
+    final String deviceModel = "Pixel 3";
```
richard-bogdan-bitrise, line 18: nit: private visibility is enough for this variable.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceNetworkTypeDataFormatterTest.java
```diff
@@ -0,0 +1,29 @@
...
+import io.bitrise.trace.data.resource.ResourceEntity;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link DeviceNetworkTypeDataFormatter}.
+ */
+public class DeviceNetworkTypeDataFormatterTest extends BaseDataFormatterTest {
+
+    final String deviceNetworkType = "WIFI";
```
richard-bogdan-bitrise, line 18: nit: private visibility is enough for this variable.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceOsDataFormatterTest.java
```diff
@@ -0,0 +1,29 @@
...
+import io.bitrise.trace.data.resource.ResourceEntity;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link DeviceOsDataFormatter}.
+ */
+public class DeviceOsDataFormatterTest extends BaseDataFormatterTest {
+
+    final String deviceOs = "11.0.0";
```
richard-bogdan-bitrise, line 18: nit: private visibility is enough for this variable.

nit: any String is perfectly ok as the value, but might worth to use a possible value: right now we get the Build.VERSION.SDK_INT (an int value) as a String, so an example could be 30, or 29, etc.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/memory/ApplicationUsedMemoryDataFormatterTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link ApplicationUsedMemoryDataFormatter}.
+ */
+public class ApplicationUsedMemoryDataFormatterTest extends BaseDataFormatterTest {
+
+    final ApplicationUsedMemoryDataFormatter formatter = new ApplicationUsedMemoryDataFormatter();
```
richard-bogdan-bitrise, line 20: nit: private visibility is enough for these member variables.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/memory/SystemMemoryDataFormatterTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link SystemMemoryDataFormatter}.
+ */
+public class SystemMemoryDataFormatterTest extends BaseDataFormatterTest {
+
+    final SystemMemoryDataFormatter formatter = new SystemMemoryDataFormatter();
```
richard-bogdan-bitrise, line 20: nit: private visibility is enough for these member variables.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/view/ApplicationStartUpDataFormatterTest.java
```diff
@@ -0,0 +1,77 @@
...
+import io.opencensus.proto.metrics.v1.Metric;
+import io.opencensus.proto.metrics.v1.MetricDescriptor;
+import io.opencensus.proto.metrics.v1.Point;
+import io.opencensus.proto.metrics.v1.TimeSeries;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for {@link ApplicationStartUpDataFormatter}
```
richard-bogdan-bitrise, line 24: nit: missing line ending dot.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/view/ApplicationStartUpDataFormatterTest.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public class ApplicationStartUpDataFormatterTest extends BaseDataFormatterTest {
+
+    final ApplicationStartUpDataFormatter formatter = new ApplicationStartUpDataFormatter();
```
richard-bogdan-bitrise, line 28: nit: private visibility is enough for these member variables.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/view/ApplicationStartUpDataFormatterTest.java
```diff
@@ -0,0 +1,77 @@
...
+    final ApplicationStartData applicationStartData = new ApplicationStartData(1234, ApplicationStartType.COLD);
+
+    @Test
+    public void formatData_notApplicationStartUp() {
+        final Data data = new Data(ApplicationStartUpDataListener.class);
+        data.setContent("kittens");
```
richard-bogdan-bitrise, line 34: ❤️ 😸

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/view/ApplicationStartUpDataFormatterTest.java
```diff
@@ -0,0 +1,77 @@
...
+        assertArrayEquals(new FormattedData[]{}, formatter.formatData(data));
+    }
+
+    @Test
+    public void formatData() {
```
richard-bogdan-bitrise, line 39: Test method is pretty long, might worth to create a different method for creating the sample/expected data. What do you think?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatterTest.java
```diff
@@ -0,0 +1,111 @@
...
+import static io.bitrise.trace.data.dto.DataValues.user;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for {@link SystemCpuDataFormatter}.
+ */
+public class SystemCpuDataFormatterTest extends BaseDataFormatterTest {
+
+    final SystemCpuDataFormatter formatter = new SystemCpuDataFormatter();
```
richard-bogdan-bitrise, line 46: nit: private visibility is enough for these member variables.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/cpu/SystemCpuDataFormatterTest.java
```diff
@@ -0,0 +1,111 @@
...
+    public void formatData() {
+        final Data data = new Data(SystemCpuUsageDataCollector.class);
+        data.setContent(cpuStat);
+
+        final FormattedData formattedData = formatter.formatData(data)[0];
+
+        final Timestamp timestamp = formattedData.getMetricEntity().getMetric()
+                .getTimeseries(0).getPoints(0).getTimestamp();
+
+        List<TimeSeries> expectedTimeSeries = new ArrayList<>(Arrays.asList(
```
richard-bogdan-bitrise, line 67: We can avoid this code duplication if the original is made package private in SystemCpuDataFormatter (with VisibleForTesting annotation), and we use it here too.

richard-bogdan-bitrise (approved):
No blockers, approved. Added some thoughts on nits and possible improvements.

As a general comment, every Data.getContent() method call can return null, so it might worth to add test cases for that too.

trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter/cpu/ApplicationCpuDataFormatter.java
daniellevass-bitrise: Interesting - hopefully having a test with real data will catch this kind of issue in the future :-)

I'll update the method on L72 to ensure it's a double not a Double 👍

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
daniellevass-bitrise: We already have this scenario covered in the [ResourceDataFormatterTest](https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/ResourceDataFormatterTest.java#L75) - as all resources *could* be null, we added that test much earlier. Does that make sense?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceCarrierDataFormatterTest.java
daniellevass-bitrise: So i thought that too - but AS gives the following warning: 
Screenshot 2021-05-10 at 15 14 47

I figured the warnings would be more annoying than the variable not being private. What do you think?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceOsDataFormatterTest.java
daniellevass-bitrise: good catch re the version number - will update 😅

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/view/ApplicationStartUpDataFormatterTest.java
daniellevass-bitrise: haha i really struggled to come up with a random string - hoped that this would do 😅

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/device/DeviceCarrierDataFormatterTest.java
richard-bogdan-bitrise: As it is used only in one place, lint is right, it can be a local variable.

daniellevass-bitrise: Have updated, and updated all the other references too (let's keep the conversation in one thread?)

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/application/ApplicationVersionCodeDataFormatterTest.java
richard-bogdan-bitrise: Looks good, thank you!

PR #191, opened 2021-04-29
chore: Updated Sonar configs

richard-bogdan-bitrise:
Updated Sonar version. Feature branches will be targeted on main, main will be not targeted. Removed caching.

APM-2913

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,7 +89,7 @@ task tagModule() {
     description = "Internal task that creates a tag for the given module"
 }
 
-String currentBranch = "main"//System.getenv("BITRISE_GIT_BRANCH")
+String currentBranch = System.getenv("BITRISE_GIT_BRANCH")
 
 sonarqube {
     properties {
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ buildscript {
 }
 
 plugins {
-    id "org.sonarqube" version "3.0"
+    id "org.sonarqube" version "3.1.1"
 }
 
 apply from: "versions.gradle"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -89,12 +89,16 @@ task tagModule() {
...
+String currentBranch = System.getenv("BITRISE_GIT_BRANCH")
+
 sonarqube {
     properties {
         property "sonar.projectKey", "bitrise-io_trace-android-sdk"
         property "sonar.organization", "bitrise-io"
         property "sonar.host.url", "https://sonarcloud.io"
-        property "sonar.branch.name", System.getenv("BITRISE_GIT_BRANCH")
-        property "sonar.branch.target", "main"
+        property "sonar.branch.name", currentBranch
```
daniellevass-bitrise, line 99: Little bit confused what the implications of this means and ` Feature branches will be targeted on main, main will be not targeted` ? as analyse_code is a manual workflow - does this mean we should only run it on feature branches, or only run it on main? Can you maybe give some more context of what this means?

richard-bogdan-bitrise: Sure, targeting means the results should contain the issues, that the given branch introduced. So let's say we have a feature branch, where we added 5 lines of code. If we target the check on the main branch, it will check what issues those 5 lines introduced, and will hide the existing issues from the main branch. More information about it can be found [here](https://sonarcloud.io/documentation/branches/overview/).

daniellevass-bitrise: ah thanks for the extra context - could you add that info to the confluence too?

daniellevass-bitrise (approved):

build.gradle
richard-bogdan-bitrise: Sure, added.

PR #192, opened 2021-04-30
chore: TraceGradlePluginFunctionalTest test dividers

richard-bogdan-bitrise:
Created FunctionalTestHelper and moved some supportive function there.
Added divider text boxes between the functional tests of
trace-gradle-plugin to help readability.

APM-2920

### DIFF
--- a/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
+++ b/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
@@ -23,6 +23,7 @@ import io.bitrise.trace.plugin.task.GenerateBuildIDTask;
 import io.bitrise.trace.plugin.task.ManifestModifierTask;
 import io.bitrise.trace.plugin.task.VerifyTraceTask;
 import io.bitrise.trace.plugin.util.FunctionalTestHelper;
+import io.bitrise.trace.plugin.util.FunctionalTestWriter;
 import io.bitrise.trace.plugin.util.TestConstants;
 
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -67,6 +68,20 @@ public class TraceGradlePluginFunctionalTest {
     private static final String buildIdFileRelativePathFormat = "/build/outputs/apk/%s/bitriseBuildId.txt";
 
     /**
+     * An instance of {@link FunctionalTestWriter} that will write the output of each test case to the console with
+     * cyan color.
+     */
+    private final FunctionalTestWriter stdOutWriter = new FunctionalTestWriter(System.out,
+            FunctionalTestWriter.PrintColor.CYAN);
+
+    /**
+     * An instance of {@link FunctionalTestWriter} that will write the error of each test case to the console with
+     * magenta color.
+     */
+    private final FunctionalTestWriter errWriter = new FunctionalTestWriter(System.err,
+            FunctionalTestWriter.PrintColor.MAGENTA);
+
+    /**
      * Sets up the test class.
      */
     @BeforeClass
@@ -101,6 +116,8 @@ public class TraceGradlePluginFunctionalTest {
         if (DELETE_TEMP_PROJECT) {
             functionalTestHelper.deleteTestProjectDir(testName);
         }
+
+        FunctionalTestWriter.reset();
     }
 
     /**
@@ -110,13 +127,8 @@ public class TraceGradlePluginFunctionalTest {
     @Test
     public void assembleDebugTraceTaskHookTest_withGradleConfig_0() {
         functionalTestHelper.setupBuildGradle(testName, 0);
+        final BuildResult buildResult = executeTaskForResult(TestConstants.ASSEMBLE_DEBUG_TASK_NAME);
 
-        final BuildResult buildResult =
-                GradleRunner.create()
-                            .withProjectDir(functionalTestHelper.getTestDir(testName))
-                            .withArguments(TestConstants.ASSEMBLE_DEBUG_TASK_NAME)
-                            .forwardOutput()
-                            .build();
         verifyDebugManifestTasks(buildResult);
         verifyDebugGenerateBuildIdTasks(buildResult);
         verifyGenerateBuildIdTasksForAssembleDebug();
@@ -129,13 +141,8 @@ public class TraceGradlePluginFunctionalTest {
     @Test
     public void assembleDebugTraceTaskHookTest_withGradleConfig_1() {
         functionalTestHelper.setupBuildGradle(testName, 1);
+        final BuildResult buildResult = executeTaskForResult(TestConstants.ASSEMBLE_DEBUG_TASK_NAME);
 
-        final BuildResult buildResult =
-                GradleRunner.create()
-                            .withProjectDir(functionalTestHelper.getTestDir(testName))
-                            .withArguments(TestConstants.ASSEMBLE_DEBUG_TASK_NAME)
-                            .forwardOutput()
-                            .build();
         verifyDebugManifestTasks(buildResult);
         verifyDebugGenerateBuildIdTasks(buildResult);
         verifyGenerateBuildIdTasksForAssembleDebug();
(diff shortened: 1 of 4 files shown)


trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
```diff
@@ -395,9 +338,11 @@ private void verifyReleaseGenerateBuildIdTasks(@NonNull final BuildResult buildR
      */
     private void verifyGenerateBuildIdTasksForBuild() {
         final File debugBuildIdFile =
-                new File(getTestDir().getPath() + "/build/outputs/apk/debug/bitriseBuildId.txt");
+                new File(functionalTestHelper.getTestDir(testName)
+                                             .getPath() + "/build/outputs/apk/debug/bitriseBuildId.txt");
```
daniellevass-bitrise, line 342: nit: should these (L342, L345, L358) be constants ? like the other file names in this class?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * Prints multiple lines with the given text in a pretty ASCII text box.
+     *
+     * @param text the text to print.
+     */
+    public void logInAsciiBox(@NonNull final String text) {
+        final int textLength = text.length();
+        final int paddingLength = 30;
```
daniellevass-bitrise, line 54: what is this padding length doing? If i understand correctly, does this mean that if the text i pass in is super long, the box widths will change? Would it be worth considering a max line length instead? so that the boxes are uniform in length and if we get one super long text, it'll automatically wrap the text? This would mean the padding might change between the borders, but maybe a min padding would be useful e.g. 5 chars either side?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
```diff
@@ -0,0 +1,298 @@
...
+     *
+     * @param testName      the given text to display.
+     * @param paddingLength the length of padding for the line, excluding starting and ending cahars.
+     * @return the line.
+     */
+    @NonNull
+    private String getTestNameLogLine(final String testName, final int paddingLength) {
+        final StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("|");
+        stringBuilder.append(concatWord(" ", paddingLength / 2));
```
daniellevass-bitrise, line 121: nit: what happens if padding length here isn't divisible by 2? I know that we're setting it up earlier to be 30, but as anyone could use this method - it might be worth just having a Math.floor here too?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * Concats the given word the given amount of times.
+     *
+     * @param word  the word to concat.
+     * @param times the number of times it should be concatenated.
+     * @return the concatenated String.
+     */
+    @NonNull
+    private String concatWord(@NonNull final String word, final int times) {
```
daniellevass-bitrise, line 137: nit: this doesn't appear to be used to concat words but single chars (L123 it's using a space char, L88 it's an equals char). Would it be more appropriate to rename to concatString instead?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
```diff
@@ -0,0 +1,298 @@
...
+    public File getTestDir(@NonNull final TestName testName) {
+        return new File(getTestDirName(testName));
+    }
+
+    /**
+     * Force-delete an existing folder and its contents, located at {@link #testProjectDir}
+     *
+     * @param testName the name of the test.
+     */
+    public void deleteTemporaryProject(@NonNull final TestName testName) {
```
daniellevass-bitrise, line 168: nit: potentially rename this to `deleteTestProjectDir` as it's the opposite of the one where it `setsUpTestDir` - see L189? possible L189 should be `setUpTestProjectDir`?

daniellevass-bitrise (approved):
Left a comment about dividers maybe having a max line length instead of being variable, and a few comments but nothing that looks blocking. Thank you for adding a comment of what the dividers looked like - this helped me understand much more easily what was happening :-)

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
richard-bogdan-bitrise: The padding length is the empty space around the test name. There is definitely a way to improve the behaviour, I decided to keep it simple because:
- this is used only for tests, no additonal functionality
- unlikely to have a super long test name (I mean it would cry for a refactor/rename if we had): let's say we would cap line length at 120 chars, it would mean ~90 chars for the testname
- if worst case scenario happens, and we have a super long name, it's functionality will be still provided, the only thing will happen it will be less good looking

Of course please let me know if I missed anything, or you would strongly prefer doing these improvements.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
richard-bogdan-bitrise: Good catch! The method concatWord expect an int as the second param. This means non divisible numbers will be rounded down. This will result in uneven line lengths (empty lines will be 1 char longer). There are different strategies to mitigate this:
1. Use only even numbers as padding, so there will be no such case. We can update the logic, to make the padding even, if it is not even.
2. Enable non even padding, this will result in asymmetric position of the test name

I would go with the first, because it is better looking. For that we can:
- indicate that use only even number as input, otherwise expect uneven box lines
- convert every input to even number (if they are not already even)
Please let me know your thoughts on this.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
richard-bogdan-bitrise: Sure, we can concat non words too, so concatString would make more sense.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
richard-bogdan-bitrise: Created a template for the common.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
daniellevass-bitrise: Hmm - sure. I think that sounds fair :) Potentially we could have a tech debt ticket to refactor and improve this in the future?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
daniellevass-bitrise: I think that sounds like a good strategy :) Perhaps we can make the current padding a constant at the top of the page with a comment with why it has to be an even number?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
daniellevass-bitrise: thank you :)

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
richard-bogdan-bitrise: Making it constant would make sense if more things would have used it, so I updated the javadoc of the given method to reflect on this.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestHelper.java
richard-bogdan-bitrise: Sure, created a ticket for it.

PR #194, opened 2021-05-03
chore: Update TraceGradlePluginFunctionalTest

richard-bogdan-bitrise:
Created FunctionalTestWriter that will change text appearance for
functional test console outputs.
Updated TraceGradlePluginFunctionalTest to use FunctionalTestWriter.
Test outputs will print text with cyan and error with magenta from now
on.

APM-2922

### DIFF
--- a/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
+++ b/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
@@ -64,17 +64,17 @@ public class TraceGradlePluginFunctionalTest {
 
     /**
      * An instance of {@link FunctionalTestWriter} that will write the output of each test case to the console with
-     * cyan color.
+     * AQUA4 color from the Bitkit.
      */
     private final FunctionalTestWriter stdOutWriter = new FunctionalTestWriter(System.out,
-            FunctionalTestWriter.PrintColor.CYAN);
+            FunctionalTestWriter.PrintColor.AQUA4);
 
     /**
      * An instance of {@link FunctionalTestWriter} that will write the error of each test case to the console with
-     * magenta color.
+     * RED4 color from the Bitkit.
      */
     private final FunctionalTestWriter errWriter = new FunctionalTestWriter(System.err,
-            FunctionalTestWriter.PrintColor.MAGENTA);
+            FunctionalTestWriter.PrintColor.RED4);
 
     /**
      * Sets up the test class.
--- a/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
+++ b/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
@@ -23,6 +23,7 @@ import io.bitrise.trace.plugin.task.GenerateBuildIDTask;
 import io.bitrise.trace.plugin.task.ManifestModifierTask;
 import io.bitrise.trace.plugin.task.VerifyTraceTask;
 import io.bitrise.trace.plugin.util.FunctionalTestHelper;
+import io.bitrise.trace.plugin.util.FunctionalTestWriter;
 import io.bitrise.trace.plugin.util.TestConstants;
 
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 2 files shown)


richard-bogdan-bitrise:
Note: branched from APM-2921.

richard-bogdan-bitrise:
Please see how it looks like in local env and on Bitrise.

Screenshot 2021-05-03 at 13 02 01
Screenshot 2021-05-03 at 13 01 16

richard-bogdan-bitrise:
Note: feel free to recommend different colors, I have no strong preference on them. The reason why I choose them. I was looking for something that is similar to the error color (red), but different, so I choose magenta, and a complementer for magenta is cyan.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
```diff
@@ -106,13 +122,8 @@ public void tearDownTest() {
     @Test
     public void assembleDebugTraceTaskHookTest_withGradleConfig_0() {
         functionalTestHelper.setupBuildGradle(testName, 0);
+        final BuildResult buildResult = executeTaskForResult(TestConstants.ASSEMBLE_DEBUG_TASK_NAME);
```
richard-bogdan-bitrise, line 125: These changes are for removing code duplication.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
```diff
@@ -97,6 +111,8 @@ public void tearDownTest() {
         if (DELETE_TEMP_PROJECT) {
             functionalTestHelper.deleteTemporaryProject(testName);
         }
+
+        FunctionalTestWriter.reset();
```
richard-bogdan-bitrise, line 115: Have to reset, otherwise the console will print with the last set color.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
```diff
@@ -0,0 +1,130 @@
...
+        final char[] escaped = concatEscapeChar(coloringCode);
+        return concatCharArrays(escaped, chars);
+    }
+
+    /**
+     * Enum class for different colors. This is based on the ANSI escape codes-
+     *
+     * @see <a href="https://en.wikipedia.org/wiki/ANSI_escape_code">https://en.wikipedia.org/wiki/ANSI_escape_code</a>
+     */
+    public enum PrintColor {
```
richard-bogdan-bitrise, line 110: Note: only 3 of them are used, but I thought we could leave the rest of them here too

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
daniellevass-bitrise: interesting - what would the overhead be if we reset after writing each text? I think this should be documented in the javadoc to ensure that if other people use it, we clear it down correctly if it does stay like this.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
daniellevass-bitrise: Moving [this conversation](https://github.com/bitrise-io/trace-android-sdk/pull/194#issuecomment-831190219) here too as they're all similar.

How much effort would it be to get the bitrise colours here from [bitkit](https://bitkit.netlify.app/documentation/materials/colors) - perhaps using red4 and aqua4 as the two colours? 

i think if we did go with the bitrise colours, it would make sense to remove the other colours we have (maybe except black and white?) - and link to bitkit to let people know where they came from? What do you think?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
richard-bogdan-bitrise: It could be reseted after each line, but it is less effective. Good point on adding this to the javadoc, updated it.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
richard-bogdan-bitrise: We can use the bitkit colors too. Updated code.
Screenshot 2021-05-04 at 13 46 20

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
daniellevass-bitrise: Yeah - i think the alternative is explicitly saying each line should be a given colour, and if you don't provide a colour, it should be default. This would have the overhead of setting up the colour and tearing down each time we write with it. Not sure what the overhead would be for this? I'm okay with the current way of working if it's now documented :)

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
daniellevass-bitrise: awesome 🎉 this would make a good 5 minute tech improvement talk btw if you're interested :)

daniellevass-bitrise (approved):

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/TraceGradlePluginFunctionalTest.java
richard-bogdan-bitrise: The behaviour is captured in the javadoc of the writer.

PR #197, opened 2021-05-06
chore: fix codecoverage for unit tests

daniellevass-bitrise:
Currently, our unit test coverage reports are not being captured and sent up. This looks to also be a problem in [the codecov tutorial](https://about.codecov.io/blog/code-coverage-for-android-development-using-kotlin-jacoco-github-actions-and-codecov/)

The fix seems to be applying the jacoco plugin to the root of the project instead of the individual project and calling the `combinedTestReportDebug` task instead. This then *only* includes the unit tests in the combined report - however the instrumentation tests are included in a different report - [fortunately the codecov step picks these all up and merges them for us so this is less of a problem](https://docs.codecov.io/docs/merging-reports).

The result looks like:
Screenshot 2021-05-06 at 10 24 12

You can check this works as this branch depends on #189  - and the code coverage for that area is now accurate https://codecov.io/gh/bitrise-io/trace-android-sdk/tree/code_cov_v3/trace-sdk/src/main/java/io/bitrise/trace/data/management/formatter

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,11 +16,10 @@ buildscript {
         mavenLocal() // To check for locally published trace-gradle-plugin artifact
         mavenCentral()
         google()
-        maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:4.1.3'
-        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.17.0-SNAPSHOT"
+        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.16.0"
         classpath files('./buildSrc/build/classes/java/main')
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,10 @@ plugins {
 apply from: "versions.gradle"
 apply plugin: "com.vanniktech.android.junit.jacoco"
 
+junitJacoco {
+    jacocoVersion = '0.8.6'
+}
+
 allprojects {
     repositories {
         mavenLocal() // For allowing the usage of local publishing and downloading them from there
--- a/build.gradle
+++ b/build.gradle
@@ -16,10 +16,11 @@ buildscript {
         mavenLocal() // To check for locally published trace-gradle-plugin artifact
         mavenCentral()
         google()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:4.1.3'
-        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.16.0"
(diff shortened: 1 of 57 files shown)


build.gradle
```diff
@@ -16,10 +16,11 @@ buildscript {
         mavenLocal() // To check for locally published trace-gradle-plugin artifact
         mavenCentral()
         google()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.3'
-        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.16.0"
+        classpath 'com.android.tools.build:gradle:4.2.0'
+        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.17.0-SNAPSHOT"
```
richard-bogdan-bitrise, line 23: Is there a reason for using a snapshot dependency? Generally I would try to avoid, as it is under constant development, and we could experience failing builds.

trace-sdk/build.gradle
```diff
@@ -78,7 +77,7 @@ dependencies {
 }
 
 junitJacoco {
-    jacocoVersion = '0.8.6'
+    jacocoVersion = '0.8.3'
```
richard-bogdan-bitrise, line 80: Why was the version of it degraded?

richard-bogdan-bitrise:
Thanks for the updates, looks good. the only thing is I would do is to avoid using snapshot dependency if there is no reason for it.

build.gradle
daniellevass-bitrise: the snapshot dependency lets us restrict the reports to just the xml or html - which in the end we're not using - i'll move it back and see if it affects the reports.

trace-sdk/build.gradle
daniellevass-bitrise: I used what's in the readme for https://github.com/vanniktech/gradle-android-junit-jacoco-plugin - i'll update it and see if it changes the reports.

richard-bogdan-bitrise (approved):

build.gradle
```diff
@@ -29,6 +29,11 @@ plugins {
 }
 
 apply from: "versions.gradle"
+apply plugin: "com.vanniktech.android.junit.jacoco"
+
+junitJacoco {
+    jacocoVersion = '0.8.4'
```
daniellevass-bitrise, line 35: this seems to be related - https://github.com/jacoco/jacoco/issues/968 - i'm not sure what happened between my PR and your PR and why that would break it. I'll add this info to the improvement ticket but i don't really want to lose another day of my life debugging this 😅

richard-bogdan-bitrise: Sure, yes, lets add a tech debt ticket to upgrade jacoco version.

PR #198, opened 2021-05-06
fix: sonar issues

daniellevass-bitrise:
Originally we started with the following issues:

Screenshot 2021-05-06 at 09 49 05

We end with the following issues:
Screenshot 2021-05-06 at 12 00 29

note: 476d0bf will cause a conflict with #189 - so this will need to be updated when that is merged first.

I fixed each issue with an individual commit and description - so it might help to read those for context too.

### DIFF
--- a/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
+++ b/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
@@ -2,7 +2,6 @@ package io.bitrise.trace.data.management;
 
 import android.content.Context;
 
-import androidx.room.Room;
 import androidx.test.platform.app.InstrumentationRegistry;
 
 import org.junit.Before;
@@ -23,9 +22,7 @@ import io.bitrise.trace.data.collector.DummyDataCollector;
 import io.bitrise.trace.data.collector.DummyDataListener;
 import io.bitrise.trace.data.dto.Data;
 import io.bitrise.trace.data.storage.DataStorage;
-import io.bitrise.trace.data.storage.MetricDao;
-import io.bitrise.trace.data.storage.TraceDataStorage;
-import io.bitrise.trace.data.storage.TraceDatabase;
+import io.bitrise.trace.data.storage.TestDataStorage;
 import io.bitrise.trace.scheduler.ServiceScheduler;
 import io.bitrise.trace.session.ApplicationSessionManager;
 
@@ -216,9 +213,7 @@ public class DataManagerInstrumentedTest {
     public void handleReceivedData_shouldHandleMetricWithoutException() {
         DataManager.getInstance(context);
         ApplicationSessionManager.getInstance().startSession();
-        final TraceDatabase traceDatabase = Room.inMemoryDatabaseBuilder(context, TraceDatabase.class).build();
-        final DataStorage traceDataStorage = TraceDataStorage.getInstance(InstrumentationRegistry.getInstrumentation().getContext());
-        traceDataStorage.setTraceDatabase(traceDatabase);
+        final DataStorage traceDataStorage = TestDataStorage.getInstance(context);
         dataManager.setDataStorage(traceDataStorage);
 
         final Data data = new Data(DataSourceType.SYSTEM_USED_MEMORY);
@@ -226,8 +221,6 @@ public class DataManagerInstrumentedTest {
         data.setContent(dummyValue);
         dataManager.handleReceivedData(data);
 
-        final MetricDao metricDao = traceDatabase.getMetricDao();
-        assertThat(metricDao.getAll().size(), is(equalTo(1)));
-        traceDatabase.close();
+        assertThat(traceDataStorage.getAllMetrics().size(), is(equalTo(1)));
     }
 }
--- a/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
+++ b/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
@@ -1,7 +1,9 @@
 package io.bitrise.trace.data.management;
 
 import android.content.Context;
+import android.os.AsyncTask;
 
+import androidx.test.annotation.UiThreadTest;
 import androidx.test.platform.app.InstrumentationRegistry;
 
 import org.junit.Before;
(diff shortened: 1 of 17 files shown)


trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
```diff
@@ -204,16 +209,25 @@ public void reset_shouldNotBeInitialised() {
...
-     * converted to a {@link io.opencensus.proto.metrics.v1.Metric} it should not throw an exception.
+     * converted to a {@link io.opencensus.proto.metrics.v1.Metric} it should not throw an exception and be saved
+     * to the database correctly.
      */
     @Test
-    @UiThreadTest
     public void handleReceivedData_shouldHandleMetricWithoutException() {
         DataManager.getInstance(context);
         ApplicationSessionManager.getInstance().startSession();
+        final TraceDatabase traceDatabase = Room.inMemoryDatabaseBuilder(context, TraceDatabase.class).build();
```
richard-bogdan-bitrise, line 219: Did you know there is [https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/androidTest/java/io/bitrise/trace/data/storage/TestDataStorage.java](https://github.com/bitrise-io/trace-android-sdk/blob/main/trace-sdk/src/androidTest/java/io/bitrise/trace/data/storage/TestDataStorage.java)? Or is it deliberate you did not use that?

trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
```diff
@@ -204,16 +209,25 @@ public void reset_shouldNotBeInitialised() {
 
     /**
      * When the {@link DataManager#handleReceivedData(Data)} is called with a valid {@link Data} that will be
-     * converted to a {@link io.opencensus.proto.metrics.v1.Metric} it should not throw an exception.
+     * converted to a {@link io.opencensus.proto.metrics.v1.Metric} it should not throw an exception and be saved
+     * to the database correctly.
      */
     @Test
-    @UiThreadTest
```
richard-bogdan-bitrise, line 210: Why did you remove this annotation?

trace-sdk/src/test/java/io/bitrise/trace/data/ULIDTest.java
```diff
@@ -36,44 +38,44 @@ public void isValid_shouldBeValid() {
         assertThat(actualValue, is(true));
     }
 
+    /**
+     * Asserts that the provided String ulid is NOT valid.
+     * @param ulid the String to validate.
```
richard-bogdan-bitrise, line 43: nit: Please use the auto-formatting.

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/BaseDataFormatter.java
```diff
@@ -10,7 +10,7 @@
  * This ensures the ApplicationSessionManager is started before which is required to be able to
  * format any data
```
richard-bogdan-bitrise, line 11: nit: I just spotted that there is no line ending dot, could you please add it there?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/BaseDataFormatter.java
```diff
@@ -10,7 +10,7 @@
...
  */
-public class BaseDataFormatterTest {
+public class BaseDataFormatter {
```
richard-bogdan-bitrise, line 13: I see your point, but I think this way it is much more confusing, because you would expect some function in it, based on it, but this is for having the common code for test classes. I would keep the original name, and instead make the class abstract, what do you think?

trace-test-application/src/androidTest/java/io/bitrise/trace/testapp/ActivityUiTest.java
```diff
@@ -25,7 +27,7 @@
     @Test
     public void activityStateTest() {
         final IndexActivityScreen indexActivityScreen = new IndexActivityScreen(uiDevice);
-
+        assertNotNull(indexActivityScreen.getButtonUiTestsLabel());
```
richard-bogdan-bitrise, line 30: I was wondering if we should add assertions for the app launch (launchTraceTestApp method) as this test is for the state of the activities of the app, not about the UI.

richard-bogdan-bitrise:
Added some questions, please check them and let me know your thoughts.

trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
daniellevass-bitrise: Because I've updated the assertion to check the item now exists in the database, room won't work on the main thread. It gives this exception:
Screenshot 2021-05-10 at 19 11 47

Is there a reason you think it should be there?

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/BaseDataFormatter.java
daniellevass-bitrise: Good catch. Added :)

trace-sdk/src/test/java/io/bitrise/trace/data/management/formatter/BaseDataFormatter.java
daniellevass-bitrise: Ooh that's a good suggestion. Will update the code and check to make sure Sonar is happy with it too.

trace-test-application/src/androidTest/java/io/bitrise/trace/testapp/ActivityUiTest.java
daniellevass-bitrise: Interesting - can you explain a bit more what you mean by that?

richard-bogdan-bitrise: So method launchTraceTestApp returns a Boolean value if the launch was successful or not, we could add assertion for that method.

trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
daniellevass-bitrise: Ah interesting problem. That explains why the test occasionally fails on CI. I've added the annotation back so that the test runs on it's own - but made the actual work on an `AsyncTask`. I'll add a comment to about why this works this way for future. Have added this in a0e7e82 .

trace-test-application/src/androidTest/java/io/bitrise/trace/testapp/ActivityUiTest.java
daniellevass-bitrise: Ah. Sure. I've updated ActivityUiTest to use this - however the other tests (ApplicationStartUpUiTest and FragmentUiTest) will still use the checking the button exists. Does that sound okay?

richard-bogdan-bitrise: Sure.

richard-bogdan-bitrise (approved):
Thanks for the updates, LGTM! 🚀

PR #199, opened 2021-05-11
tests: improve coverage for data collectors

daniellevass-bitrise:
Have added as many tests as I could to increase the code coverage. 

Some lines will be impossible to cover, others would take a long time to write a test for so I prioritised getting 80+% coverage. 

Also optimised to get more unit tests than instrumented tests to hopefully reduce the reliance on emulators.

Before:
Screenshot 2021-05-11 at 14 17 27

Current:
Screenshot 2021-05-18 at 15 37 41

Updated 20/05/2021
Screenshot 2021-05-20 at 16 29 03


trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/application/ApplicationVersionCodeDataCollectorTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        final Data expectedData = new Data(ApplicationVersionCodeDataCollector.class);
+        expectedData.setContent(String.valueOf(versionCode));
+
+        assertEquals(expectedData, collector.collectData());
+
+        ConfigurationManager.reset();
+    }
+
+    @Test(expected = Resources.NotFoundException.class)
```
daniellevass-bitrise, line 45: I'm not sure this is the best behaviour - this would cause a crash if the customer didn't have a version_code or version_name. Would a better solution to catch this exception in the collectors and make it an empty string instead?

trace-sdk/src/main/java/io/bitrise/trace/data/collector/view/ApplicationStartUpDataListener.java
```diff
@@ -132,12 +138,8 @@ public void onActivityResumed(@NonNull final Activity activity) {
      * @see
      * <a href="https://developer.android.com/topic/performance/vitals/launch-time">https://developer.android.com/topic/performance/vitals/launch-time</a>
      */
-    @Nullable
+    @NonNull
     private synchronized ApplicationStartData getApplicationStartData() {
-        if (!isApplicationStart) {
```
daniellevass-bitrise, line 137: removed these lines of code because this method is private, and only one place that calls it - and that is surrounded by the check already - so this check was unnecessary.

daniellevass-bitrise:
Could we delete the TraceLogCollector - it's not being used for anything currently?

trace-sdk/src/main/java/io/bitrise/trace/data/collector/memory/MemoryDataCollector.java
```diff
@@ -9,6 +9,6 @@
 
     @Override
     public long getIntervalMs() {
-        return 15 * 1000;
```
daniellevass-bitrise, line 12: this seemed like redundant maths - i hope it's okay i simplified it :)

trace-sdk/src/main/java/io/bitrise/trace/data/dto/FragmentData.java
```diff
@@ -117,4 +118,16 @@ public String toString() {
                 ", spanId='" + spanId + '\'' +
                 '}';
     }
+
+    @Override
+    public boolean equals(@Nullable final Object obj) {
```
daniellevass-bitrise, line 123: added a bunch of the equal operators to a few data classes, as the unit tests were giving false negatives if i asked to assert if two objects were equals.

trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/device/DeviceCarrierDataCollectorInstrumentedTest.java
```diff
@@ -2,26 +2,28 @@
...
-     * {@link Data} should not be {@code null}.
+     * {@link Data} should match the expected data.
      */
     @Test
     public void collectData_contentShouldBeNotNull() {
-        final DeviceCarrierDataCollector deviceCarrierDataCollector = new DeviceCarrierDataCollector(context);
-        final String actualValue = (String) deviceCarrierDataCollector.collectData().getContent();
-        assertThat(actualValue, is(notNullValue()));
+        final Data expectedData = new Data(DeviceCarrierDataCollector.class);
+        expectedData.setContent("Android");
```
richard-bogdan-bitrise, line 26: Test will fail if tested on real devices, or if the given emulated device does not have a sim card. I think we should keep the test as is. A new test can be added, if we check that the device has a sim card and is an emulated device.

trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/device/DeviceLocaleDataCollectorInstrumentedTest.java
```diff
@@ -1,26 +1,57 @@
...
      */
     @Test
     public void collectData_contentShouldBeNotNull() {
-        final DeviceLocaleDataCollector deviceLocaleDataCollector = new DeviceLocaleDataCollector(context);
-        assertThat(deviceLocaleDataCollector.collectData().getContent(), is(notNullValue()));
+        assertThat(collector.collectData().getContent(), is(notNullValue()));
     }
+
+    @Test
+    public void collectData_mockData() {
```
richard-bogdan-bitrise, line 38: This test verifies that we are getting the locale data from resources/configuration. It would fail if we get the locale from a different source. Do we need such test?

trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollectorInstrumentedTest.java
```diff
@@ -1,28 +1,40 @@
...
-        assertThat(actualValue, is(notNullValue()));
+    @SdkSuppress(maxSdkVersion = 28)
+    public void collectData_pre29() {
+        final DeviceNetworkTypeDataCollector collector = new DeviceNetworkTypeDataCollector(context);
+        assertNetworkCollectedData(collector.collectData(), "UNKNOWN");
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 29)
+    public void collectData_post29() {
```
richard-bogdan-bitrise, line 29: nit: post stands fro after, but 29 is also included.

trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
```diff
@@ -77,16 +77,20 @@ public static synchronized ConfigurationManager getInstance() {
     /**
      * Creates an instance of the Configuration Manager for debugging uses only. You can pass a token
      * that should be used for the trace addon configuration.
-     * @param token the token that should be used to work with trace.
+     *
+     * @param token  the token that should be used to work with trace.
+     * @param config items to add to the configurationMap e.g. VERSION_CODE
```
richard-bogdan-bitrise, line 82: nit: missing line ending dot.

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceCarrierDataCollector.java
```diff
@@ -34,7 +36,8 @@ public DeviceCarrierDataCollector(@NonNull final Context context) {
     @Override
     public Data collectData() {
         final Data data = new Data(this);
-        data.setContent(getDeviceCarrier(context));
+        data.setContent(getDeviceCarrier((TelephonyManager) context.getSystemService(
```
richard-bogdan-bitrise, line 39: Why did you move this code from the method below?

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollector.java
```diff
@@ -130,17 +135,15 @@ private static String getNetworkTypeFromNetworkCapabilities(@NonNull final Conne
     /**
      * Gets the type of the cellular network with the help of the TelephonyManager.
      *
-     * @param context the Android Context.
+     * @param telephonyManager the telephonyManager from the context
```
richard-bogdan-bitrise, line 138: nit: missing line ending dot.

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollector.java
```diff
@@ -80,7 +83,7 @@ private static String getDeviceNetworkType(@NonNull final Context context) {
     @RequiresPermission(android.Manifest.permission.ACCESS_NETWORK_STATE)
     @NonNull
     @Deprecated
-    private static String getNetworkTypeDataFromNetworkInfo(@NonNull final ConnectivityManager connectivityManager) {
+    static String getNetworkTypeDataFromNetworkInfo(@NonNull final ConnectivityManager connectivityManager) {
```
richard-bogdan-bitrise, line 86: L86, L110, L146: did you change the visibility of them just for the tests to use?

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class ApplicationVersionCodeDataCollectorTest {
+
+    private final static int versionCode = 123;
+    private final Context mockContext = Mockito.mock(Context.class);
+    private final ApplicationVersionCodeDataCollector collector = new ApplicationVersionCodeDataCollector(
+            mockContext);
+
+    @Test
+    public void collectData_configurationManagerInitialised() {
```
richard-bogdan-bitrise, line 30: Similarly to [this](https://github.com/bitrise-io/trace-android-sdk/pull/199/files#r635145991) comment, this verifies that we get the data from the ConfigurationManager. What happens if configurationManager returns null or a non valid value during operation? This test will still pass, but we will have issues. I think we should check we get a valid version code instead.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import io.bitrise.trace.BuildConfig;
+import io.bitrise.trace.configuration.ConfigurationManager;
+import io.bitrise.trace.data.collector.application.ApplicationVersionCodeDataCollector;
+import io.bitrise.trace.data.dto.Data;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for {@link ApplicationVersionCodeDataCollector}
```
richard-bogdan-bitrise, line 20: Missing line ending dot.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionNameDataCollectorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import io.bitrise.trace.BuildConfig;
+import io.bitrise.trace.configuration.ConfigurationManager;
+import io.bitrise.trace.data.collector.application.ApplicationVersionNameDataCollector;
+import io.bitrise.trace.data.dto.Data;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for {@link ApplicationVersionNameDataCollector}
```
richard-bogdan-bitrise, line 20: nit: missing line ending dot.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionNameDataCollectorTest.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class ApplicationVersionNameDataCollectorTest {
+
+    private final static String versionName = "1.2.3";
+    private final Context mockContext = Mockito.mock(Context.class);
+    private final ApplicationVersionNameDataCollector collector = new ApplicationVersionNameDataCollector(
+            mockContext);
+
+    @Test
+    public void collectData_configurationManagerInitialised() {
```
richard-bogdan-bitrise, line 30: Please see [this](https://github.com/bitrise-io/trace-android-sdk/pull/199/files#r635170192).

trace-sdk/src/test/java/io/bitrise/trace/data/collector/cpu/CpuStatTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Test
+    public void equals_differentCpuStat() {
+        final CpuUsageData.CpuStat differentCpuStat = new CpuUsageData.CpuStat(
+                1, 2, 3, 4, 5, 6, 7, 80 );
+        assertFalse(cpuStat.equals(differentCpuStat));
+    }
+
+    @Test
+    public void hashCode_shouldMatch() {
```
richard-bogdan-bitrise, line 40: Following the logic above, shouldn't we test instead, that hash code of the same object is equal? (instead of a magic number)

trace-sdk/src/test/java/io/bitrise/trace/data/collector/network/okhttp/OkHttpDataListenerTest.java
```diff
@@ -169,4 +179,41 @@ public void createNetworkData_ShouldCreateCorrectNetworkData() {
                                                                   .setStart(DUMMY_START_TIME).setEnd(DUMMY_END_TIME);
         assertThat(actualValue, is(expectedValue));
     }
+
+    @Test
+    public void constructor() {
+        final OkHttpDataListener listener = new OkHttpDataListener(mockContext);
+        assertNotNull(listener);
```
richard-bogdan-bitrise, line 186: I do not think constructor calls can return null.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ApplicationForegroundStateDataListenerTest.java
```diff
@@ -0,0 +1,38 @@
...
+import android.content.Context;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertFalse;
+
+/**
+ * Unit tests for {@link ApplicationForegroundStateDataListener}
```
richard-bogdan-bitrise, line 13: missing line ending dot.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ApplicationStartUpDataListenerTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+/**
+ * Unit tests for {@link ApplicationStartUpDataListener}
```
richard-bogdan-bitrise, line 19: missing line ending dot.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unit tests for {@link ActivityStateDataListener}.
+ */
+public class ActivityStateDataListenerTest {
+
+    final Context mockContext = Mockito.mock(Context.class);
+    final ActivityStateDataListener listener = new ActivityStateDataListener(mockContext);
+    final Activity mockActivity = Mockito.mock(Activity.class);
+
+    @Test
+    public void onActivityCreated_notActive() {
```
richard-bogdan-bitrise, line 23: not Active tests already tested with activityMap_ShouldNotContainActivityWhenCollectingStopped in ActivityStateDataListenerInstrumentedTest

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ApplicationForegroundStateDataListenerTest.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class ApplicationForegroundStateDataListenerTest {
```
richard-bogdan-bitrise, line 15: Similarly to [this](ApplicationForegroundStateDataListenerInstrumentedTest), this is already tested implicitly in ApplicationForegroundStateDataListenerInstrumentedTest, except for onStop.

richard-bogdan-bitrise:
The following do not have coverage, but it would have been good:
[https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/cpu/SystemCpuUsageDataCollector.java#L184](https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/cpu/SystemCpuUsageDataCollector.java#L184)
[https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/cpu/SystemCpuUsageDataCollector.java#L232](https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/cpu/SystemCpuUsageDataCollector.java#L232)
[https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/network/okhttp/TraceOkHttpInterceptor.java#L19](https://github.com/bitrise-io/trace-android-sdk/blob/APM-2830_collector_tests/trace-sdk/src/main/java/io/bitrise/trace/data/collector/network/okhttp/TraceOkHttpInterceptor.java#L19)

trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/device/DeviceLocaleDataCollectorInstrumentedTest.java
daniellevass-bitrise: I think the purpose of this was to make sure it matched a given input e.g. if the locale was set to "en" the data that came out should also be "en" and we hadn't changed it at all. But I agree, it maybe isn't as useful as the others - so it could be removed.

trace-sdk/src/androidTest/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollectorInstrumentedTest.java
daniellevass-bitrise: updated to read `collectData_sdk29AndAfter` and `collectData_preSdk29` - does that make it clearer?

trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
daniellevass-bitrise: added the dot.

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceCarrierDataCollector.java
daniellevass-bitrise: It was easier to provide a mocked TelephonyManager to the method for the unit tests, than to mock giving it to the context to then get used. If that makes sense?

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollector.java
daniellevass-bitrise: put the dot back.

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceNetworkTypeDataCollector.java
daniellevass-bitrise: Yeah. I thought I had added the annotation but it must've got lost somewhere - apologies. I'll put the annotation over these changes.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
daniellevass-bitrise: As this is a unit test, we don't actually have a version_code or version_name available which is why i'm creating a debug configurationManager with those set.

 I had added a comment (which I later deleted) that said something about if the configurationManager doesn't have the version_code or version_name available, it will crash the customers application. I had instrumented tests for this - but i decided they weren't actually very useful to assume a crash. But I didn't think refactoring the behaviour here made sense either.

My suggestion would be - we should re-prioritise APM-2424 (higher priority?) - we should update that ticket to also add tests for the ConfigurationManager around finding things that aren't available and have an agreed expected behaviour (empty string or null, not a crash).  If that makes sense?

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
daniellevass-bitrise: Added the dot.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionNameDataCollectorTest.java
daniellevass-bitrise: Added a dot.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionNameDataCollectorTest.java
daniellevass-bitrise: Wrote the reply in that thread - to keep the conversation in one place.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/cpu/CpuStatTest.java
daniellevass-bitrise: Sure - I can update my code to test the same object provides the same hash code :)

trace-sdk/src/test/java/io/bitrise/trace/data/collector/network/okhttp/OkHttpDataListenerTest.java
daniellevass-bitrise: maybe this was an eager test - will remove.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
daniellevass-bitrise: I think if you look at the coverage report on main for that it's displaying those rows as red - https://codecov.io/gh/bitrise-io/trace-android-sdk/src/main/trace-sdk/src/main/java/io/bitrise/trace/data/collector/view/ActivityStateDataListener.java - so i think theses tests still are valid.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ApplicationForegroundStateDataListenerTest.java
daniellevass-bitrise: Similarly, the coverage report informs they're being missed - https://codecov.io/gh/bitrise-io/trace-android-sdk/src/main/trace-sdk/src/main/java/io/bitrise/trace/data/collector/view/ApplicationForegroundStateDataListener.java - i think these tests are valid.

trace-sdk/src/main/java/io/bitrise/trace/data/collector/device/DeviceCarrierDataCollector.java
richard-bogdan-bitrise: Cool, thanks for the explanation.

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
richard-bogdan-bitrise: Sure, we can add this to that ticket. But I believe these tests just add some comfort by increasing the coverage numbers, but have no real value.

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
richard-bogdan-bitrise: I think I got the reason. So this assertion is redundant, but that line is not covered, because stopCollecting also unregisters the listener, so the onStop method of the listener will be not called, because the traceActivityLifecycleTracker will not forward it.
So I was considering if those isActive checks are needed or not, because there is no use case when it is not active, but registered. From tech perspective, we cannot foretell if the usage of those method changes in the future (and how), so I believe, for safety reasons we should keep the checks, and therefore the tests are needed too.
On the other hand, a really similar case you mentioned [here](https://github.com/bitrise-io/trace-android-sdk/pull/199#discussion_r633488933), because of current usage the (!isApplicationStart) check is redundant, but applying the same principle, I think we should keep that check there too. Thoughts?

trace-sdk/src/main/java/io/bitrise/trace/data/collector/memory/MemoryDataCollector.java
richard-bogdan-bitrise: I usually keep the components separate, to make it more understandable, but in this simple case when we just multiply with 1000 I agree with the simplification. 👍

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
daniellevass-bitrise: Thanks for the comment Richard. I think I agree with what you're saying, this class (and the other) would maybe worth a bit of a refactor bearing that information in mind. My intention for this ticket wasn't to try and refactor any of the existing code (acceptance criteria is only to increase current code coverage). Let's add this to the android catchup agenda on Monday to discuss how we can improve this going forward? Would that work for you?

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
daniellevass-bitrise: So i think these tests have value - they're comfort checking that given an input we get an expected output as there's some processing in the middle (it would be easy to set a different content for example, maybe get the NAME/CODE properties mixed up). Unit tests are great for us so we can be confident that nothing in that process changes in the future.

richard-bogdan-bitrise: I believe it would make more sense if we would check it with real data, because this will never fail, if the config item name changes from "VERSION_CODE" to something else in the ConfigurationManager, but not in the Collector. But agree, it meets the acceptance criteria...

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
richard-bogdan-bitrise: So you removed the check in one place, but kept in the other and added test cases for it. This is not about refactoring, but inconsistency in your PR.

richard-bogdan-bitrise:
Thanks for the additional test cases! 🎉

richard-bogdan-bitrise (approved):

trace-sdk/src/test/java/io/bitrise/trace/data/application/ApplicationVersionCodeDataCollectorTest.java
daniellevass-bitrise: So i think there's an interesting scenario if we get integrated into someone else's library - libraries don't get version_codes anymore - so this would currently cause a crash. IMO we need to deal with this scenario better - defaulting to an empty string / or unknown instead of a crash :)

trace-sdk/src/test/java/io/bitrise/trace/data/collector/view/ActivityStateDataListenerTest.java
daniellevass-bitrise: I don't think I've removed any code - i definitely didn't intend to remove any checks. Can you point to example of this so I can put it back? There are some line deletions because I moved the collection creation to class visibility, instead of each test creating it, but that should be the only change to existing tests here.

PR #200, opened 2021-05-12
chore: Improve FunctionalTestWriter

richard-bogdan-bitrise:
FunctionalTestWriter does not require reset from now on.

APM-2922

### DIFF
--- a/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
+++ b/trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
@@ -1,23 +1,24 @@
 package io.bitrise.trace.plugin.util;
 
 import androidx.annotation.NonNull;
-
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.util.Arrays;
 
 /**
- * Custom implementation of the {@link PrintWriter} class, to add different coloring to the output and some extra
- * indentation. Please note that adding the coloring will stay until changed. To reset it to the default please use
- * the {@link #reset()} method. The coloring is based on the ANSI escape codes.
+ * Custom implementation of the {@link PrintWriter} class, to add different coloring to the
+ * output and some extra
+ * indentation. The coloring is based on the ANSI escape codes.
  *
- * @see <a href="https://en.wikipedia.org/wiki/ANSI_escape_code">https://en.wikipedia.org/wiki/ANSI_escape_code</a>
+ * @see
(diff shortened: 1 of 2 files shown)


richard-bogdan-bitrise:
So it was bugging me, that we have to reset the writer after the functional tests have run, because it is possible to forget that, so I updated a the code a bit, now it does not requires the reset.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
```diff
@@ -17,7 +18,7 @@
     private static final char[] indentation = new char[]{' ', ' ', ' '};
     private static final char[] escapeChar = new char[]{27};
     private final char[] coloringCode;
-    private final int extraLength;
+    private final char[] coloringEnd = concatEscapeChar("[39;49m".toCharArray());
```
daniellevass-bitrise, line 21: what does `[39;49m` mean here?

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
```diff
@@ -106,7 +91,7 @@ public void write(@NonNull final char[] chars, final int i, final int i1) {
 
     /**
      * Inserts to the start the coloring chars to a given array of chars. Please note that adding the coloring will
-     * stay until changed. To reset it to the default please use the {@link #reset()} method.
+     * stay until changed.
```
daniellevass-bitrise, line 94: nit: not sure this javadoc makes sense "the coloring will stay until changed" - i think the "until changed" isn't helpful - how do we change it? Perhaps something like "until a new color is requested, or the functional tests complete"?

daniellevass-bitrise (approved):
Thanks for taking the time to improve this! 

Left two comments about making things clearer to others who read this code :)

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
richard-bogdan-bitrise: Thanks for spotting it, removed them.

trace-gradle-plugin/src/functionalTest/java/io/bitrise/trace/plugin/util/FunctionalTestWriter.java
richard-bogdan-bitrise: That is the required ANSI code to end the coloring.

PR #209, opened 2021-05-25
tests: Improve coverage around scheduling

daniellevass-bitrise:
Before:
Screenshot 2021-05-21 at 11 43 42

After:
Screenshot 2021-05-25 at 14 05 20


trace-sdk/src/main/java/io/bitrise/trace/scheduler/ScheduledJobService.java
```diff
@@ -1,49 +0,0 @@
-package io.bitrise.trace.scheduler;
```
daniellevass-bitrise, line 1: I removed this file as it looks like it wasn't being used by anything. Do you see any issue with this?

trace-sdk/src/main/java/io/bitrise/trace/scheduler/ServiceScheduler.java
```diff
@@ -29,6 +30,10 @@
   @NonNull
   private final Class<? extends JobService> jobService;
 
+  @VisibleForTesting
+  @NonNull
+  ComponentName componentName;
```
daniellevass-bitrise, line 35: this ComponentName caused issues whilst testing - I've made a class property so that I can override it in the test.

trace-sdk/src/test/java/io/bitrise/trace/scheduler/ExecutorSchedulerTest.java
```diff
@@ -0,0 +1,112 @@
...
+  public void scheduleDelayed_intervalGreaterThanZero() {
+    final ExecutorScheduler executorScheduler =
+        new ExecutorScheduler(mockContext, mockRunnable, initialDelay, 200);
+
+    executorScheduler.scheduleDelayed(initialDelay);
+
+    assertNotNull(executorScheduler.scheduledFuture);
+    assertTrue(executorScheduler.scheduledFuture.getDelay(TimeUnit.MILLISECONDS)
+        > (initialDelay - 100));
+    // we cannot get the delay immediately, several milliseconds will have passed.
```
daniellevass-bitrise, line 56: Not sure how you feel about this - we can't get the exact time that was set because the time will have elapsed - i figured if it's within 100ms of what we set it's probably okay. If it's instant the time will be zero or negative (the delay has passed) so it's easy to compare that? What do you think?

richard-bogdan-bitrise: I have no strong opinion on this.

PR #210, opened 2021-05-27
tests: Increase coverage around network package

daniellevass-bitrise:
Increased code coverage around the network package.

Before:
Screenshot 2021-05-21 at 11 44 17

After:

Screenshot 2021-06-02 at 11 48 43


trace-sdk/src/main/java/io/bitrise/trace/network/MetricSender.java
```diff
@@ -129,11 +130,8 @@ static int countHeaderComparisonDifference(@NonNull final MetricRequest request,
   public void onSuccess() {
     TraceLog.i(LogMessageConstants.METRIC_SENT_SUCCESSFULLY);
     getExecutor().execute(() -> {
-      if (!getMetricEntityList().isEmpty()) {
```
daniellevass-bitrise, line 132: i don't think this can actually happen - we check when we send if we have metrics to send - i don't think it would be changed before the request has been completed?

trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java
```diff
@@ -116,7 +117,7 @@ static synchronized NetworkCommunicator getCommunicator() {
   @NonNull
   public static Gson getGson() {
     return new GsonBuilder()
-        .setFieldNamingStrategy(new SnakeCaseStrategy())
```
daniellevass-bitrise, line 119: removed the SnakeCaseStrategy class and usage here in favour of the FieldNamingPolicy. LOWER_CASE_WITH_UNDERSCORES - as it means this is code we don't have to maintain (or test). 

However, we're not actually using this line as all the info that gets sent goes through the `ProtoTypeAdapter` which has it's serializationFormat set to CaseFormat.LOWER_UNDERSCORE.  

I think it's maybe useful to keep if we add more classes in the future that we send, we don't have to change the field naming policy - what do you think?

trace-sdk/src/main/java/io/bitrise/trace/network/adapters/PointAdapter.java
```diff
@@ -37,29 +37,31 @@ public Point deserialize(@NonNull final JsonElement json, @NonNull final Type ty
...
-              srcPoint.get(PROPERTY_VALUE), SummaryValue.class);
-          pointBuilder.setSummaryValue(summaryValue);
-          break;
-        case VALUE_NOT_SET:
-          // we deliberately do not deserialize values that were not set.
-          break;
-        default:
-          // nop.
-          break;
+      if (valueCase != null) {
```
daniellevass-bitrise, line 40: this was a warning in android studio - and i accidentally wrote a test that caught it. My assumption was `Point.ValueCase.forNumber(srcValueCase.getAsInt());` would return the number even if it couldn't find it, and then it would hit the default block in the switch - however it looks like the method will return a null instead - so this needs to be checked it's not null.

daniellevass-bitrise:
GitHub wouldn't let me leave this as a comment in the relevant code base but I think we may have a bug in MetricSender and TraceSender in the `send` method: we have a `try` block which calls `onSendFinished` once if it is stopped, and once if the network request is invalid - however we also have a `finally` which also calls `onSendFinished` as well. This means there are scenarios where this method gets called twice. [here is the area i'm talking about in MetricSender](https://github.com/bitrise-io/trace-android-sdk/pull/210/files#diff-2c32687197486b6b0e1746fd78860d67204ae6f079a6e074fece6bc549526d24R99-R122).

Is that a problem? Or should we remove these calls in the try block. 

I tried to write these up in a test but because it's in an executor and all the other methods have a lot of dependencies it's been very hard. Will come back to this if I have time later in the week.

trace-sdk/src/main/java/io/bitrise/trace/network/DataSender.java
```diff
@@ -84,7 +88,7 @@
    * @param networkRequest the given NetworkRequest.
    * @return {@code true} if it is valid, {@code false} otherwise.
    */
-  static boolean validateTraceRequest(@Nullable final NetworkRequest networkRequest) {
```
daniellevass-bitrise, line 87: refactored this method because it's not just validating trace requests, it validates all network requests.

trace-sdk/src/main/java/io/bitrise/trace/network/NetworkClient.java
richard-bogdan-bitrise: Lets keep it.

richard-bogdan-bitrise:
> GitHub wouldn't let me leave this as a comment in the relevant code base but I think we may have a bug in MetricSender and TraceSender in the `send` method: we have a `try` block which calls `onSendFinished` once if it is stopped, and once if the network request is invalid - however we also have a `finally` which also calls `onSendFinished` as well. This means there are scenarios where this method gets called twice. [here is the area i'm talking about in MetricSender](https://github.com/bitrise-io/trace-android-sdk/pull/210/files#diff-2c32687197486b6b0e1746fd78860d67204ae6f079a6e074fece6bc549526d24R99-R122).
> 
> Is that a problem? Or should we remove these calls in the try block.
> 
> I tried to write these up in a test but because it's in an executor and all the other methods have a lot of dependencies it's been very hard. Will come back to this if I have time later in the week.

I think it should be fixed.

richard-bogdan-bitrise:
> > GitHub wouldn't let me leave this as a comment in the relevant code base but I think we may have a bug in MetricSender and TraceSender in the `send` method: we have a `try` block which calls `onSendFinished` once if it is stopped, and once if the network request is invalid - however we also have a `finally` which also calls `onSendFinished` as well. This means there are scenarios where this method gets called twice. [here is the area i'm talking about in MetricSender](https://github.com/bitrise-io/trace-android-sdk/pull/210/files#diff-2c32687197486b6b0e1746fd78860d67204ae6f079a6e074fece6bc549526d24R99-R122).
> > Is that a problem? Or should we remove these calls in the try block.
> > I tried to write these up in a test but because it's in an executor and all the other methods have a lot of dependencies it's been very hard. Will come back to this if I have time later in the week.
> 
> I think it should be fixed.

Note: the fix mght not belong to this scope, so we can create a new task for it.  What would you prefer?

richard-bogdan-bitrise (approved):

daniellevass-bitrise:
> > > GitHub wouldn't let me leave this as a comment in the relevant code base but I think we may have a bug in MetricSender and TraceSender in the `send` method: we have a `try` block which calls `onSendFinished` once if it is stopped, and once if the network request is invalid - however we also have a `finally` which also calls `onSendFinished` as well. This means there are scenarios where this method gets called twice. [here is the area i'm talking about in MetricSender](https://github.com/bitrise-io/trace-android-sdk/pull/210/files#diff-2c32687197486b6b0e1746fd78860d67204ae6f079a6e074fece6bc549526d24R99-R122).
> > > Is that a problem? Or should we remove these calls in the try block.
> > > I tried to write these up in a test but because it's in an executor and all the other methods have a lot of dependencies it's been very hard. Will come back to this if I have time later in the week.
> > 
> > 
> > I think it should be fixed.
> 
> Note: the fix might not belong to this scope, so we can create a new task for it. What would you prefer?

I think as it's a pretty simple fix (deleting 4 lines of code) i'm happy to add it to this PR - if that's okay with you?

richard-bogdan-bitrise:
Sure, thanks!

PR #211, opened 2021-06-02
tests: Increase coverage for core data

daniellevass-bitrise:
Class | Coverage Before | Coverage After
--- | --- | ---
DataManager | 79.25% | 91.82%
TraceActivityLifecycleTracer | 98.67% | no change
TraceSdk | 8.77% | 69.49%
 ApplicationSessionManager | 70.59% | 100.00%

Mostly improved coverage around the DataManager and TraceSdk.

There are still some areas that are more difficult to test e.g. the TraceSdk init method - as we can't inject mocks or change  the underlying objects such as a ConfigurationManager - and that causes issues.

### DIFF
--- a/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
+++ b/trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
@@ -7,15 +7,12 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
 
 import android.content.Context;
 import android.os.AsyncTask;
 import androidx.test.annotation.UiThreadTest;
 import androidx.test.platform.app.InstrumentationRegistry;
 import io.bitrise.trace.configuration.ConfigurationManager;
-import io.bitrise.trace.data.collector.DataCollector;
-import io.bitrise.trace.data.collector.DataListener;
 import io.bitrise.trace.data.collector.DataSourceType;
 import io.bitrise.trace.data.collector.DummyDataCollector;
 import io.bitrise.trace.data.collector.DummyDataListener;
@@ -24,10 +21,6 @@ import io.bitrise.trace.data.storage.DataStorage;
 import io.bitrise.trace.data.storage.TestDataStorage;
 import io.bitrise.trace.scheduler.ServiceScheduler;
 import io.bitrise.trace.session.ApplicationSessionManager;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Set;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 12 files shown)


trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
```diff
@@ -112,8 +114,14 @@ public static synchronized void init(@NonNull final Context context) {
     getInstance();
     initialised = true;
     configurationMap = new HashMap<>();
-    importConfigurationFromBuildConfig(context);
-    importConfigurationFromResValues(context);
+
+    try {
```
daniellevass-bitrise, line 118: this regularly causes crashes when unit/integration testing without the token set - i don't think these should ever crash anything - but should definitely show an error instead. What do you think?

trace-sdk/src/main/java/io/bitrise/trace/data/management/DataManager.java
```diff
@@ -286,20 +305,21 @@ public void startSending(@NonNull final Context context) {
   /**
    * Stops the sending of any pending requests to the server. These request will be dismissed and
    * will not be sent. No future requests will be sent unless {@link #startSending} is called again.
-   *
-   * @param context the Android Context.
    */
-  public void stopSending(@NonNull final Context context) {
```
daniellevass-bitrise, line 292: I refactored stopSending a little bit :

Previously - stopSending would create the metricServiceScheduler and traceServiceScheduler if they didn't exist, to then call cancel - then startSending would check if they were null (which they wouldn't ever be).

I've now made it, so that it checks if the metricServiceScheduler or traceServiceScheduler is not null before posting a log (I think this was misleading to people who read the logs) - it then only calls cancel on them if they were created.  This means the creation of objects stays in start which i think makes more sense.

 Does that sound okay?

trace-sdk/src/main/java/io/bitrise/trace/utils/log/TraceLog.java
```diff
@@ -31,10 +31,6 @@ private TraceLog() {
    */
   public static synchronized Logger makeAndroidLogger() {
     synchronized (lock) {
-      if (logger != null) {
```
daniellevass-bitrise, line 34: I refactored the logger a little bit too - previously if you tried to make the logger a different kind of logger after it was already initialised, it wouldn't change e.g. i created an AndroidLogger but now I wanted a SilentLogger - i would expect it to change. 

Does that sound okay?

trace-sdk/src/main/java/io/bitrise/trace/session/ApplicationSessionManager.java
```diff
@@ -76,4 +77,14 @@ public synchronized void stopSession() {
   public Session getActiveSession() {
     return session;
   }
+
+  @VisibleForTesting
+  static synchronized boolean isSessionManagerActive() {
```
daniellevass-bitrise, line 82: Added these two booleans for the tests - so that i could verify that the reset made both things null. Does that make sense?

trace-sdk/src/androidTest/java/io/bitrise/trace/data/management/DataManagerInstrumentedTest.java
```diff
@@ -83,80 +65,24 @@ public void setUp() {
...
   public void stopSending_shouldCallCancelAll() {
-    dataManager.stopSending(context);
+    dataManager.stopSending();
 
     verify(dataManager.metricServiceScheduler, times(1)).cancelAll();
     verify(dataManager.traceServiceScheduler, times(1)).cancelAll();
   }
 
   /**
-   * Test when the Data collection in started in the DataManager, the number of active
```
daniellevass-bitrise, line 98: all these deletions are because there are now unit tests (cheaper vs integration) that cover the same test functions.

trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
richard-bogdan-bitrise: Makes sense.

trace-sdk/src/main/java/io/bitrise/trace/data/management/DataManager.java
richard-bogdan-bitrise: Sure.

trace-sdk/src/main/java/io/bitrise/trace/utils/log/TraceLog.java
richard-bogdan-bitrise: Yes.

trace-sdk/src/main/java/io/bitrise/trace/session/ApplicationSessionManager.java
richard-bogdan-bitrise: Okay.

richard-bogdan-bitrise (approved):

PR #214, opened 2021-06-09
feature: Crash Listener and Formatter

daniellevass-bitrise:
Created a crash listener and formatter.

The TraceExceptionDataListener extends BaseDataListener however, it doesn't call onDataCollected with a Data object as we're not using an opencensus style object so i've created it's own CrashData. 

This means the DataManager now has a new method to handleReceivedCrash which uses ExceptionDataFormatter to format the data.

Note: Using a feature branch for this work in case we need to make any bug fixes onto main in the interim

### DIFF
--- a/trace-sdk/src/main/java/io/bitrise/trace/data/collector/crash/TraceExceptionDataListener.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/data/collector/crash/TraceExceptionDataListener.java
@@ -1,0 +1,55 @@
+package io.bitrise.trace.data.collector.crash;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+import io.bitrise.trace.data.collector.BaseDataListener;
+import io.bitrise.trace.data.dto.CrashData;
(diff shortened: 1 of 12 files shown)


trace-sdk/src/main/java/io/bitrise/trace/data/collector/crash/TraceExceptionDataListener.java
```diff
@@ -0,0 +1,55 @@
...
+import androidx.annotation.VisibleForTesting;
+import io.bitrise.trace.data.collector.BaseDataListener;
+import io.bitrise.trace.data.dto.CrashData;
+import io.bitrise.trace.data.management.formatter.crash.ExceptionDataFormatter;
+
+/**
+ * Listens for any unhandled exceptions that get thrown and passes them to the
+ * {@link ExceptionDataFormatter}.
+ */
+public class TraceExceptionDataListener extends BaseDataListener
```
daniellevass-bitrise, line 14: what do you think of `TraceExceptionDataListener` instead of `ExceptionDataListener`?

richard-bogdan-bitrise (approved):

PR #216, opened 2021-06-11
test: Add ExceptionActivity

richard-bogdan-bitrise:
trace-test-application will have a test screen, where we can generate sample crashes.

APM-3120

### DIFF
--- a/trace-test-application/src/main/java/io/bitrise/trace/testapp/exception/ExceptionActivity.java
+++ b/trace-test-application/src/main/java/io/bitrise/trace/testapp/exception/ExceptionActivity.java
@@ -1,0 +1,127 @@
+package io.bitrise.trace.testapp.exception;
+
+import android.os.Bundle;
+import android.widget.Button;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.SwitchCompat;
+import io.bitrise.trace.session.ApplicationSessionManager;
+import io.bitrise.trace.session.SessionManager;
+import io.bitrise.trace.testapp.R;
+import okhttp3.OkHttpClient;
+
+/**
+ * Activity that will throw exceptions causing the app to crash.
+ */
+public class ExceptionActivity extends AppCompatActivity {
+
+  /**
+   * The {@link SessionManager} to control session starts and ends.
+   */
+  private final SessionManager sessionManager = ApplicationSessionManager.getInstance();
+
+  @Override
+  protected void onCreate(@Nullable final Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_exception);
+    setupView();
+  }
+
+  /**
+   * Sets up the initial state of the view.
+   */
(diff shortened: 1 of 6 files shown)


trace-test-application/src/main/res/layout/activity_exception.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    android:padding="20dp"
+    tools:context=".IndexActivity">
```
daniellevass-bitrise, line 8: nit: this should be .ExceptionActivity?

trace-test-application/src/main/res/layout/activity_exception.xml
```diff
@@ -0,0 +1,45 @@
...
+        android:id="@+id/btn_exception_main_thread"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Throw exception on main thread" />
+
+    <Button
+        android:id="@+id/btn_exception_thread"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Throw  exception on thread" />
```
daniellevass-bitrise, line 44: nit you have two space characters between "throw" and "exception".

trace-test-application/src/main/java/io/bitrise/trace/testapp/exception/ExceptionActivity.java
```diff
@@ -0,0 +1,127 @@
...
+   * Throws an exception in the main thread. Should be called from the main thread.
+   */
+  private void throwExceptionInMainThread() {
+    throwException("Main thread is not happy!");
+  }
+
+  /**
+   * Creates a new thread and throws an exception.
+   */
+  private void throwExceptionInThread() {
```
daniellevass-bitrise, line 101: nit: I'd rename this as `throwExceptionInNewThread` and replace the button text too - it's a bit confusing as you have Main Thread, and then Thread (of which Main Thread is a Thread) - explicitly adding this is a "new" thread would help readability.

trace-test-application/src/main/java/io/bitrise/trace/testapp/exception/ExceptionActivity.java
```diff
@@ -0,0 +1,127 @@
...
+    new Thread() {
+      @Override
+      public void run() {
+        super.run();
+        throwException(String.format("Thread with name %s is not happy",
```
daniellevass-bitrise, line 106: nit: all your other exception messages end with a `!` character - this should either end with a full stop or other ending character.

daniellevass-bitrise (approved):
Left a couple of nits to look at. Thank you for your work on this 🎉

trace-test-application/src/main/res/layout/activity_exception.xml
richard-bogdan-bitrise: It is not needed, removed.

PR #219, opened 2021-06-22
fix: datamanager resource bug

daniellevass-bitrise:
It looks like I introduced a bug whilst writing the unit tests for this class, this meant resources are no longer being saved to the data storage. This PR ensures they now get saved, and has a test for them too.

I also optimised the resource collection - these are items that do not change during the application lifecycle so don't need to be regularly collected. Collecting them once when Trace starts is fine.

### DIFF
--- a/trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
+++ b/trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
@@ -257,17 +257,27 @@ public class ConfigurationManager {
    * @return the DataCollectors.
    */
   @NonNull
-  public Set<DataCollector> getDataCollectors(@NonNull final Context context) {
+  public Set<DataCollector> getRecurringDataCollectors(@NonNull final Context context) {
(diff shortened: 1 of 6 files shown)


trace-sdk/src/main/java/io/bitrise/trace/configuration/ConfigurationManager.java
```diff
@@ -257,17 +257,27 @@ public boolean isAppDebugBuild() {
...
 
+  /**
+   * Creates a new set of {@link DataCollector}s that only need to collect their data once during
+   * the lifecycle of the application e.g. Application version codes.
+   *
+   * @param context the Android Context.
+   * @return the DataCollectors.
+   */
+  @NonNull
+  public Set<DataCollector> getSingleDataCollectors(@NonNull final Context context) {
```
daniellevass-bitrise, line 277: i didn't have a better name for "something that needs to collect data only once" :/ Do you have any other suggestion?

daniellevass-bitrise:
This has been approved via alternative means.

